PR #4, opened 2019-10-15
✨Data seeder+other bits

blargly:
Data seeder for ref data. This posts data to the new APIs.
Also some changes due to warnings, refactor of some DTOs and Controller end points due to back and forth debates on what the correct response should be for DELETE


src/Common/Common/Data/ReferenceData/AnnualStatisticGroup.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+
+namespace Common.Data.ReferenceData
+{
+    /// <summary>
+    /// Ref Data. A group which contains a collection of
```
dan-giddins: Is this comment meant to be like this?

src/RefDataLoader/RefDataLoader/Config/ApiSettings.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Collections.Generic;
+
+namespace Config
+{
+    /// <summary>
+    /// Settings related to the Ref Data API
+    /// </summary>
+    public class ApiSettings
+    {
+        public string Baseuri { get; set; }
```
dan-giddins: is this capitalisation okay?

beforan: `BaseUri` would be better really

src/Common/Common/Data/ReferenceData/AnnualStatisticGroup.cs
blargly: Amended

src/RefDataLoader/RefDataLoader/Config/ApiSettings.cs
blargly: Amended

PR #18, opened 2019-11-11
⚡Core: Minor Frontend refactor

beforan:
- Added ESLint with `create-react-app` settings so Server Page frontend is in line with the ClientApp
- Webpack runs ESLint
- Fixed ESLint issues
- `routes.js` is now just a route definition file
- `main.js` now does the Server Page React bootstrapping
- `App.js` now defines the Server Page React App entrypoint

### DIFF
--- a/src/Directory/Directory/Pages/Account/components/RegisterForm.js
+++ b/src/Directory/Directory/Pages/Account/components/RegisterForm.js
@@ -1,7 +1,7 @@
 import React, { useState } from "react";
 import { Formik, Form, Field } from "formik";
 import { postObjectAsFormData, constants } from "js-forms";
-import { Button, Stack, Box, Flex, Grid, SimpleGrid } from "@chakra-ui/core";
+import { Button, Stack, Box, SimpleGrid } from "@chakra-ui/core";
 import valSchema from "./register-form-validation";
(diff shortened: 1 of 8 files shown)


src/Directory/Directory/Pages/Account/components/RegisterForm.js
```diff
@@ -1,7 +1,7 @@
 import React, { useState } from "react";
 import { Formik, Form, Field } from "formik";
 import { postObjectAsFormData, constants } from "js-forms";
-import { Button, Stack, Box, Flex, Grid, SimpleGrid } from "@chakra-ui/core";
+import { Button, Stack, Box, SimpleGrid } from "@chakra-ui/core";
```
Rhino12343: not sure if its related but you've removed box from the passwordrequirementsinfo.js file and its still used here.

beforan: that's because it's actually used in this file, but wasn't in the other 👍

Rhino12343: yeah sorry I just realised that, I got myself confused.
Hardly looked at a computer for the last 9 days haha

PR #24, opened 2020-07-31
Publication Class Library and Example App

Flemingjp:
Created the publications class library, with an example implemented web app that can be ran locally.

- SQL Entities and Migration for a Publication Entity
- Service to connect to the Directory and call its `/api/Biobanks` endpoint to get list of Organisations
- Service to connect to the Europe PMC API to search for publications by ID and Organisation name
- Service to read/write publications to a SQL database
- Background service to populate database with publications


Flemingjp:
Superseded by #35

PR #35, opened 2020-08-18
Publication azure

JosephLavagna:
Azure Function Publications Work
- HTTP Trigger Function(EPMC Function) pulls from external publication API for specific biobank and outputs a JSON result to be consumed by the directory. 
- Schedule (every 24hrs) Function (Batch Function) pulls all biobanks from test, pulls the associated publications from the API and pushes the result to an Azure sql db.
- Works alongside the class library in same solution.
- Pipelines build the functions app into a zip package, builds a sql script and then deploys those two to azure respectively.
- Need to migrate to biobank azure subscription


Flemingjp:
Superceeds #34 

> Created the publications class library, with an example implemented web app that can be ran locally.
> 
> - SQL Entities and Migration for a Publication Entity
> - Service to connect to the Directory and call its `/api/Biobanks` endpoint to get list of Organisations
> - Service to connect to the Europe PMC API to search for publications by ID and Organisation name
> - Service to read/write publications to a SQL database
> - Background service to populate database with publications
>

src/Publications/Publications/Publications.csproj
```diff
@@ -0,0 +1,39 @@
...
+    <Compile Remove="Migrations\20200811121721_testAzure.Designer.cs" />
+    <Compile Remove="Migrations\20200814100414_testinggg.cs" />
+    <Compile Remove="Migrations\20200814100414_testinggg.Designer.cs" />
+    <Compile Remove="Migrations\20200817091834_testingParser.cs" />
+    <Compile Remove="Migrations\20200817091834_testingParser.Designer.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="MatthiWare.CommandLineParser" Version="0.3.0" />
+    <PackageReference Include="Microsoft.AspNetCore.WebUtilities" Version="2.2.0" />
```
beforan, line 28: What are we using WebUtilities for?

If we *are* using it, we should use the same version as the target framework - `3.1` not `2.2`.

src/Publications/Publications/Services/BiobankWebService.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Publications.Services
+{
+    public class BiobankWebService : IBiobankService, IDisposable
+    {
+        private readonly HttpClient _client;
+
+        public BiobankWebService(IConfiguration configuration)
+        {
+            _client = new HttpClient
```
beforan, line 17: can we dependency inject this? using `IHttpClientFactory`? https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests

src/Publications/Publications/Services/Dto/PublicationDTO.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+    public class EPMCSearchResult
+    {
+        [JsonProperty("nextCursorMark")]
+        public string Cursor { get; set; }
+
+        [JsonProperty("resultList")]
+        public Results Results { get; set; }
+
+        public List<PublicationDTO> Publications
```
beforan, line 17: expression bodies for readonly getters pls

`public List<PublicationDTO> Publications => Results.Publications;`

src/Publications/Publications/Services/Dto/PublicationDTO.cs
```diff
@@ -0,0 +1,50 @@
...
+        }
+    }
+
+    public class Results
+    {
+        [JsonProperty("result")]
+        public List<PublicationDTO> Publications { get; set; }
+    }
+
+    public class PublicationDTO
```
beforan, line 29: CamelCase Acronyms (arguably ones longer than 2 letters) say this should be `PublicationsDto`.

Resharper or other format tooling would complain ;)

https://stackoverflow.com/questions/15526107/acronyms-in-camelcase

src/Publications/Publications/Services/Dto/PublicationDTO.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        [JsonProperty("id")]
+        public int Id { get; set; }
+
+        [JsonProperty("title")]
```
beforan, line 34: does Newtonsoft / System.Text.Json not automatically convert between `PascalCase` (C#) / `camelCase` (JSON)? rendering some of these annotations unnecessary?

src/Publications/Publications/Services/EMPCWebService.cs
```diff
@@ -0,0 +1,87 @@
...
+using System.Runtime.ExceptionServices;
+using Microsoft.VisualBasic.CompilerServices;
+using Microsoft.AspNetCore.WebUtilities;
+using System.Linq;
+using Publications.Services.Contracts;
+using Microsoft.Extensions.Configuration;
+
+namespace Publications
+{
+     public class EMPCWebService : IEPMCService, IDisposable
```
beforan, line 18: is it `EMPC` or `EPMC`? (interface and class don't match).

Also previous notes on casing acronyms.

src/Publications/Publications/Services/EMPCWebService.cs
```diff
@@ -0,0 +1,87 @@
...
+    {
+
+        private readonly HttpClient _client;
+        
+        public EMPCWebService(IConfiguration configuration)
+        {
+            _client = new HttpClient
```
beforan, line 25: dependency inject from IHttpClientFactory

src/Publications/Publications/Services/EMPCWebService.cs
```diff
@@ -0,0 +1,87 @@
...
+        public async Task<List<PublicationDTO>> GetOrganisationPublications(string biobank)
+        {
+            List<PublicationDTO> publications = new List<PublicationDTO>();
+            
+            string query = $"ACK_FUND:\"{biobank}\"";
+            string nextCursor = "*";
+            string currentCursor;
+
+            //API Pagination loop - will stop once current cursor and next cursor are equal (no more records)
+            do
```
beforan, line 45: snap! don't see many `do ... while`s in C# these days. totally appropriate use though 👍

src/Publications/Publications/Services/EMPCWebService.cs
```diff
@@ -0,0 +1,87 @@
...
+        public async Task<PublicationDTO> GetPublicationById(int publicationId)
+        {
+            return (await Search($"{publicationId}")).Publications.FirstOrDefault();
+        }
+
+        public async Task<List<PublicationDTO>> GetOrganisationPublications(string biobank)
+        {
+            List<PublicationDTO> publications = new List<PublicationDTO>();
+            
+            string query = $"ACK_FUND:\"{biobank}\"";
```
beforan, line 40: we use `var` for assigned variables.

beforan: explicit types appear in various places in this file, not just here ;)

src/Publications/Publications/Services/Hosted/FetchPublicationsService.cs
```diff
@@ -0,0 +1,55 @@
...
+            _logger = logger;
+            _scopeFactory = scopeFactory;
+        }
+
+        public async Task StartAsync(CancellationToken cancellationToken)
+        {
+            using (var scope = _scopeFactory.CreateScope())
+            {
+                // DI of required services
+                var publicationService = scope.ServiceProvider.GetRequiredService<IPublicationService>();
```
beforan, line 33: this is Service Location, not DI.
DI is preferable, and IHostedService supports constructor injection (that's how you got the ScopeFactory).

[This](https://github.com/biobankinguk/biobankinguk/tree/master/src/RefDataLoader/RefDataLoader) uses IHostedService and DI's its dependencies.

src/Publications/Publications/Services/PublicationService.cs
```diff
@@ -0,0 +1,82 @@
...
+        private PublicationDbContext _ctx;
+
+        public PublicationService(PublicationDbContext ctx)
+        {
+            _ctx = ctx;
+        }
+
+        public async Task AddOrganisationPublications(string organisationName, IEnumerable<PublicationDTO> publications)
+        {
+            var existingPublications = _ctx.Publications.Where(x => x.Organisation.Equals(organisationName));
```
beforan, line 26: don't see a lot of `.Equals`. I think `==` is immediately clearer if there's no reason to explicitly use `.Equals` (after some research there are very few reasons, especially with known-to-be strings)

src/Publications/PublicationsAzureFunctions/Properties/serviceDependencies.PublicationsAzureFunctionsNew - Zip Deploy.json
```diff
@@ -0,0 +1,10 @@
+{
+  "dependencies": {
```
beforan, line 2: what are these files?

src/Publications/PublicationsAzureFunctions/PublicationsAzureFunctions.csproj
```diff
@@ -0,0 +1,34 @@
...
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="3.1.6" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore.Tools" Version="3.1.6">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="Microsoft.Extensions.Http" Version="3.1.6" />
+    <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="3.0.9" />
```
beforan, line 20: should this be `3.1.x` to match the target framework?

src/Publications/PublicationsAzureFunctions/README.md
```diff
@@ -0,0 +1,5 @@
+
+
```
beforan, line 2: more to come here? are we using GH actions?

src/Publications/PublicationMockApp/Properties/launchSettings.json
```diff
@@ -0,0 +1,30 @@
...
+  "iisSettings": {
+    "windowsAuthentication": false,
+    "anonymousAuthentication": true,
+    "iisExpress": {
+      "applicationUrl": "http://localhost:54938",
+      "sslPort": 44305
+    }
+  },
+  "profiles": {
+    "IIS Express": {
```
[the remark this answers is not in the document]
Flemingjp: Resolved -> Removed PublicationMockApp

src/Publications/Publications/Migrations/PublicationDbContextModelSnapshot.cs
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Publications.Migrations
+{
+    [DbContext(typeof(PublicationDbContext))]
+    partial class PublicationDbContextModelSnapshot : ModelSnapshot
+    {
+        protected override void BuildModel(ModelBuilder modelBuilder)
+        {
+#pragma warning disable 612, 618
+            modelBuilder
```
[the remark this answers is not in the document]
Flemingjp: Snapshot was generated during initial migration - what is the alternative and why would it have use ModelBuilder this time?

src/Publications/Publications/Entities/Publication.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace Publications.Entities
+{
+    public class Publication
+    { 
+        [Key]
+        public int InternalId { get; set; }
+
+        public int PublicationId { get; set; }
+
+        public string Organisation { get; set; }
```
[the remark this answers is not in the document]
Flemingjp: In the strictest sense, all of them except for DOI, as some publications wont have an assigned DOI. Pratically though some maybe passable as null - eg Authors/Journal

src/Publications/Publications/Publications.csproj
Flemingjp: Used to parse a url from a base url and a dictionary of query parameters. 
Usage => 
`QueryHelpers.AddQueryString("webservices/rest/search", parameters);`

Will update, and remove if really needed - more verbose, manual solutions exist

src/Publications/PublicationsAzureFunctions/PublicationsAzureFunctions.csproj
Flemingjp: Latest version appears to be 3.0.9 
https://www.nuget.org/packages/Microsoft.NET.Sdk.Functions/

src/Publications/PublicationsAzureFunctions/README.md
Flemingjp: More to come here - idea is to add how to deploy using the azure pipelines, as it requires some azure portal and devops setting up. 

Should this be done before merge, or later? It can be cleaned up to remove the confusion over Github Actions.

src/Publications/PublicationsAzureFunctions/Properties/serviceDependencies.PublicationsAzureFunctionsNew - Zip Deploy.json
Flemingjp: Appear to be configuration for the azure function - the code is deployed as a .zip to a storage instance, and the azure web service connects to that storage instance on launch/reload.

src/Publications/Publications/Services/PublicationService.cs
Flemingjp: Fair comment and good to know, slipped into Java conventions :P

src/Publications/Publications/Entities/Publication.cs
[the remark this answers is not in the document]
beforan: so I would suggest one of the following:
- Enable [Nullable Reference Types](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-reference-types) on the project, and then make the nullable strings explicitly nullable using `string?`
  - then EF will base the DB column nullable on the C# types.
- Annotate any strings that shouldn't be nullable with `[Required]`
  - then EF will make DB columns non-nullable for Required properties.

Either is fine. `[Required]` is older but more common, Nullable Reference types are newer but turning them on may have other side effects that you don't want to resolve right now.

src/Publications/PublicationsAzureFunctions/Properties/serviceDependencies.PublicationsAzureFunctionsNew - Zip Deploy.json
beforan: Do they all refer to current Azure environments? I'm suspicious of the filenames, e.g. `PublicationsAzureFunctionsNew`.

Any that were made for testing and no longer used should be removed.

src/Publications/PublicationsAzureFunctions/README.md
beforan: I think it can be done post merge as long as it gets done, say, this sprint? I'd consider documenting this work part of completing it, ideally, else in 5 years time we'll be wishing we'd written something down here.

src/Publications/Publications/Publications.csproj
beforan: Ugh it probably only comes included with `Microsoft.NET.Sdk.Web` projects, which this ain't. Fair enough.

src/Publications/Publications/Migrations/PublicationDbContextModelSnapshot.cs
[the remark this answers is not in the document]
beforan: my bad, didn't notice this was the snapshot file! 🤭

src/Publications/PublicationsAzureFunctions/Properties/serviceDependencies.PublicationsAzureFunctionsNew - Zip Deploy.json
Flemingjp: These are apparently development specific, except for the servicesDependencies.json. I've deleted these files and added a new rule to the gitignore for `./*.local.json`

beforan (approved):

PR #37, opened 2020-08-28
Publications migration to Azure Key Vault

Flemingjp:
Migrated storing database credentials to the Azure Key Vault. The azure pipelines is now configured to fetch these credentials during the Deploy stage, and are subsequently used to deploy the database migration. The pipeline is also used to apply KeyVault references to the Azure Functions App. This allows the Azure Functions App to access the credentials from the functions app settings at runtime - without needing to access the KeyVault via the codebase.

Also migrated the api urls to the app settings from the connection strings - as recommended by Microsoft in their documentation.

Also changed the EpmcFunction to use the correct service `EpmcService => PublicationService`

Added README explaining how to configure the service


.azure/pipelines/publications.yml
```diff
@@ -8,8 +8,13 @@ trigger:
     - src/Publications/*
 
 variables:
+  ## Azure Resources
+  resourceGroupName: 'biobankinguk-services-test'
```
blargly, line 12: The resource group should be 'biobankinguk-directory-test' as that's the existing one we're using for the test environment.

.azure/pipelines/publications.yml
```diff
@@ -8,8 +8,13 @@ trigger:
...
+  keyVaultName: 'biobanking-services-test'
```
blargly, line 13: The existing keyvault in the above resource Group is 'directoryTestKeyvault'

.azure/pipelines/publications.yml
```diff
@@ -8,8 +8,13 @@ trigger:
...
+  dbServerName: 'biobankinguk-services-test.database.windows.net'
```
blargly, line 14: Again, may as well use an existing resource: 'biobankinguk-directory-test'

beforan (approved):

src/Publications/PublicationsAzureFunctions/Startup.cs
```diff
@@ -20,25 +20,29 @@ namespace PublicationsAzureFunctions
     class Startup : FunctionsStartup
     {
         private IConfiguration _configuration;
+        
         public override void Configure(IFunctionsHostBuilder builder)
         {
-            var config = builder.Services.BuildServiceProvider()
+            _configuration = builder.Services.BuildServiceProvider()
```
beforan, line 26: for this, can we build only the configuration early, rather than the entire service tree?

```suggestion
            _configuration = builder.ConfigurationBuilder.Build();
```

beforan (changes requested):

src/Publications/Publications/Services/EpmcWebService.cs
```diff
@@ -23,7 +23,7 @@ public class EpmcWebService : IEpmcService, IDisposable
         public EpmcWebService(IHttpClientFactory httpClientFactory, IConfiguration configuration)
```
beforan, line 23: All the services you use `HttpClientFactory` in don't need to dispose the client anymore.

beforan (changes requested):

src/Publications/PublicationsAzureFunctions/Startup.cs
Flemingjp: `IFunctionsHostBuilder` only exposes the Services property - maybe another way to do this?

blargly (approved):

src/Publications/PublicationsAzureFunctions/Startup.cs
beforan: yeah, my bad. misread the docs

beforan (approved):

PR #40, opened 2020-10-19
Writing publication data to the Directory DB

shakisparki:
Writing publication data to the Directory DB instead of separate publications db


src/Publications/Publications/Repositories/GenericEFRepository.cs
```diff
@@ -0,0 +1,152 @@
+using Microsoft.EntityFrameworkCore;
```
beforan, line 1: Please god no do not use this code. This class needs to die.

This is a new app with EF Core, just use the DbContext directly, since that is in effect an implementation of the repository pattern.

src/Publications/Publications/Repositories/IGenericEFRepository.cs
```diff
@@ -0,0 +1,39 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Threading.Tasks;
+
+namespace Directory.Data.Repositories
+{
+    public interface IGenericEFRepository<TEntity>
```
beforan, line 9: Obviously this can go too

src/Publications/Publications/PublicationDbContext.cs
```diff
@@ -17,6 +18,7 @@ public class PublicationDbContext : DbContext
         public PublicationDbContext(DbContextOptions<PublicationDbContext> options) : base(options) { }
 
         public DbSet<Publication> Publications { get; set; }
+        public DbSet<Organisation> Organisations { get; set; }
```
beforan, line 21: If this is read-only it can (should?) be a `DbQuery<T>` not a `DbSet`

src/Publications/Publications/Entities/Organisation.cs
```diff
@@ -0,0 +1,56 @@
+using System;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+
+namespace Directory.Data.Entities
+{
+    public class Organisation
```
beforan, line 7: I'm quite uncomfortable with duplicating this entity over here in this codebase.

The right thing to do would be to put entities in a class library and reference that everywhere that uses those entities. This is close to the fact the Data stuff in the Directory is already in a class library, but I recognise the need to have the entities separate from the Contexts/Migrations, especially in view of the fact that's EF6 and everything newer is likely to be EF Core.

If, as in an ideal world, this code and the directory code were in the same repo, a class library of entities would be a no brainer.

Sadly, due to the different repos currently (and probably for a while longer), we'd have to publish a nuget package of the entities to share them, or use git submodules. Definitely faffy.

@blargly thoughts? Is the duplication a necessary evil until the directory code can live in the same repo? Or is publishing a package or using submodules worth it to avoid it?
How else do we ensure that schema changes to Organisation in the directory don't break publications because it's schema is elsewhere.
Discussions around SubmissionAPI/Directory alignment may affect decisions on this too.

src/Publications/Publications/PublicationDbContext.cs
beforan: sorry that's out of date. `DbQuery<>` is deprecated.

src/Publications/Publications/Entities/Organisation.cs
beforan: Also, I notice this PR removes migrations. So is the Directory now responsible for the Publication entity? In which case I have the same concerns about that.

beforan: Chatted with @blargly we'll leave the duplication for now, as we now have something of a roadmap for the Directory to resolve these issues with regards to the API as well as the other dependent services.

There will be work to undo this duplication on that roadmap, so it should hopefully be relatively shortlived, given the renewed focus on the API.

beforan (dismissed):

src/Publications/Publications/Services/PublicationService.cs
```diff
@@ -66,7 +66,7 @@ public async Task AddOrganisationPublications(string organisationName, IEnumerab
         public async Task<IEnumerable<PublicationDto>> GetOrganisationPublications(string organisationName)
         {
```
Flemingjp, line 67: Is this method used anymore? If not remove it?

Flemingjp (changes requested):

Flemingjp (approved):

beforan (approved):

PR #43, opened 2020-11-12
Move ref data controller to Api Controller

shakisparki:
Move refdata controllers to Api controllers


src/Directory/Biobanks.Web/ApiControllers/AccessConditionsController.cs
```diff
@@ -0,0 +1,162 @@
...
+
+            //Everything went A-OK!
+            return Json(new
+            {
+                msg = $"The access condition \"{model.Description}\" was deleted successfully.",
+                type = FeedbackMessageType.Success
+            });
+        }
+
+        private IHttpActionResult JsonModelInvalidResponse(ModelStateDictionary state)
```
shakisparki, line 150: better to move this to separate controller class?

beforan: It's more like a helper method but that they all use. Either chuck it in some kind of utils class, or create a base api controller with it on, I guess.

beforan:
This PR needs to wait for the directory codebase migration to GitHub to complete. then we can rebase and go from there.

Flemingjp:
Closed in favour of its rebase

PR #50, opened 2020-11-20
Move RefData Controller to Api Controller

Flemingjp:
This is a rebase of #43 onto the new Open Source Directory branch.

~~**This PR currently merges into `oss-directory` to make the changes clearer - at time of merge it most likely wants to merge into `master`**~~


Flemingjp:
Reminded of https://github.com/biobankinguk/biobankinguk/pull/43#discussion_r525047679

shakisparki:
> Reminder of [#43 (comment)](https://github.com/biobankinguk/biobankinguk/pull/43#discussion_r525047679)

Resolved

src/Directory/Biobanks.Web/ApiControllers/AccessConditionsController.cs
```diff
@@ -0,0 +1,150 @@
...
+            return Json(new
+            {
+                success = true,
+                name = model.Description,
+                redirect = $"AddAccessConditionSuccess?name={model.Description}"
+            });
+        }
+
+        [HttpPost]
+        public async Task<IHttpActionResult> EditAccessConditionAjax(AccessConditionModel model)
```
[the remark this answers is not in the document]
beforan: also the signature and routing needs changing to take the `id` in the url, to make it RESTful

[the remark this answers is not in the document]
beforan: and the verb should be PUT

src/Directory/Biobanks.Web/Scripts/ADAC/adac-access-conditions.js
```diff
@@ -61,7 +61,7 @@ function AdacAccessConditionViewModel() {
 
         // Get Action Type
         var action = _this.modal.mode().toLowerCase();
-        var url = `${action}AccessConditionAjax`;
+        var url = `/api/AccessConditions/${action}AccessConditionAjax`;
```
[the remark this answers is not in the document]
beforan: also, can we put the url in one place, probably in a class property

src/Directory/Biobanks.Web/Scripts/ADAC/adac-access-conditions.js
```diff
@@ -128,7 +128,7 @@ $(function () {
         var triggerRow = diff.filter(row => row.node == edit.triggerRow.node())[0];
 
         //AJAX Update
-        $.post("EditAccessConditionAjax?sortOnly=true",
+        $.post("/api/AccessConditions/EditAccessConditionAjax?sortOnly=true",
```
beforan, line 131: can we have a sort endpoint?

something like:
- `/api/<Resource>/sort`
- `/api/<Resource>/<id>/move`
- `/api/<Resource>/<id>/move/<new position>`

src/Directory/Biobanks.Web/Scripts/ADAC/adac-access-conditions.js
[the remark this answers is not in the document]
beforan: the behaviour change needs to be as follows:

instead of POSTing the model (id and description?) to either `Add<X>` or `Edit<X>`, we need to:

- If "Add", POST to `/api/<Resource>` with the payload as it is now, I guess
- If "Edit", PUT at `/api/<Resource>/<id>` with the payload as it is now. But this'll mean getting the id and using it in the url

src/Directory/Biobanks.Web/App_Start/WebApiConfig.cs
```diff
@@ -14,6 +14,10 @@ public class WebApiConfig
 
         public static void Register(HttpConfiguration config)
         {
+            config.Routes.MapHttpRoute(
+                name: "DefaultApiWithAction",
+                routeTemplate: "api/{controller}/{action}"
```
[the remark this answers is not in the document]
beforan: If we add a `sort` endpoint as suggested below, there will be action endpoints on otherwise RESTful resources. I would suggest using [Attribute-based Routing](https://docs.microsoft.com/en-us/aspnet/web-api/overview/web-api-routing-and-actions/attribute-routing-in-web-api-2), to cover cases that stray from the default route template of `api/<controller>/<id>?`

It's clearer, more flexible and closer to ASP.NET Core's default conventions.

src/Directory/Biobanks.Web/Views/ADAC/AnnualStatisticGroups.cshtml
```diff
@@ -93,9 +93,9 @@
             </div>
 
             <form id="modal-annual-statistic-group-form"
-                  data-update-action="@Url.Action("EditAnnualStatisticGroupAjax")"
+                  data-update-action="/api/AnnualStatisticGroups/EditAnnualStatisticGroupAjax"
```
beforan, line 96: wait, the Views have data attributes with the URLs but the JS never uses them? Amazing.

That's the other option; put the URL here (only one URL though now of course, because it's just the resource endpoint `/api/<Resource>` in `data-resource-url` or something.

What are the Success URLs about? feedback? Where are the routes for those? still on `ADACController`? That doesn't seem right...

src/Directory/Biobanks.Web/Web.config
```diff
@@ -54,6 +54,10 @@
    
     <!-- Environment -->
     <add key="EnvironmentName" value="Development" />
+
+    <!--Ref Data Api-->
+    <add key="RefDataApiUrl" value="https://localhost:44300/"/>
```
beforan, line 59: what needs this? This app never uses an external API for ref data.

src/Directory/Services/BiobankWriteService.cs
```diff
@@ -1503,7 +1503,7 @@ public async Task DeletePreservationTypeAsync(PreservationType preservationType)
         #region RefData: Material Type
         public async Task DeleteMaterialTypeAsync(MaterialType materialType)
         {
-            await _materialTypeRepository.DeleteAsync(materialType);
+            await _materialTypeRepository.DeleteAsync(materialType.MaterialTypeId);
```
beforan, line 1506: is this a bug fix?

beforan (changes requested):

beforan:
The point of this is not to just move the methods from `ADACController`, though that is a good move anyway, but to make a REST API for ref data that the other systems can also use as necessary, as well as the Directory UI. So it should follow modern API standards in the way it behaves (e.g. HTTP VERBs, not action POST endpoints).

Also I really hate that there's like 30 almost identical `.js` files... but we can't really confidently fix that nicely til we fix frontend bundling. So it'll have to wait. But we really should have one `ref-data-admin-view.js` and you just give it slightly differently shaped payloads and different endpoints (read from the view).

src/Directory/Biobanks.Web/ApiControllers/AccessConditionsController.cs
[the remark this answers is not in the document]
shakisparki: @beforan For the PUT controller, model already contains an Id, so which signature is best here?

EditAccessConditionAjax(int id, [FromBody]string Description, [FromBody]int SortOrder)

EditAccessConditionAjax(int id, AccessConditionModel model)

EditAccessConditionAjax(AccessConditionModel model)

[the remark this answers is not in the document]
beforan: I was thinking `(int id, Model model)`

I realise we technically duplicate the `id` then. Technically we should probably use a different model that doesn't include `id` but i'm not super fussed. The `id` param is the one to use though, not the model one

src/Directory/Biobanks.Web/Views/ADAC/AnnualStatisticGroups.cshtml
shakisparki: Yes, the URLs are for feedback. Where do you expect them to be? 
Isn't the feedback given dependent on the application not the API?
and if not in ADACController, where else could it reside? in the ApiController?

beforan (changes requested):

src/Directory/Biobanks.Web/Views/ADAC/AnnualStatisticGroups.cshtml
beforan: I've reviewed the way this currently works and i see that it is needed. I can't think of a better place to put it right now; definitely not the ApiController, so we'll leave it.

src/Directory/Biobanks.Web/Scripts/ADAC/adac-access-conditions.js
```diff
@@ -9,7 +9,7 @@ function AccessCondition(id, description, sortOrder) {
 
 function AccessConditionModal(id, description, sortOrder) {
     this.modalModeAdd = "Add";
-    this.modalModeEdit = "Edit";
+    this.modalModeEdit = "Update";
```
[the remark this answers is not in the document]
shakisparki: some already used update, and others edit. Simply easier for me to make them similar for easier modification

src/Directory/Services/BiobankWriteService.cs
shakisparki: didn't seem to work as before

src/Directory/Biobanks.Web/ApiControllers/CollectionPointController.cs
```diff
@@ -0,0 +1,169 @@
...
+            {
+                success = true,
+                name = model.Description,
+            });
+        }
+
+        [HttpDelete]
+        [Route("")]
+        public async Task<IHttpActionResult> DeleteCollectionPoint(int id)
+        {
```
Flemingjp, line 119: Should this endpoint just be named `Delete`

src/Directory/Biobanks.Web/ApiControllers/RegistrationReasonController.cs
```diff
@@ -0,0 +1,140 @@
...
+            {
+                success = true,
+                name = model.Description
+            });
+        }
+
+        [HttpPut]
+        [Route("{id}")]
+        public async Task<IHttpActionResult> Put(int id, Models.Shared.RegistrationReasonModel model)
+        {
```
Flemingjp, line 81: Is it cleaner and/or clearer to use `using RegistrationReasonModel = Biobanks.Web.Models.Shared.RegistrationReasonModel`

src/Directory/Biobanks.Web/ApiControllers/ServiceOfferingController.cs
```diff
@@ -0,0 +1,168 @@
...
+                success = true,
+                name = model.Name
+            });
+
+        }
+
+        [HttpPut]
+        [Route("{id}")]
+        public async Task<IHttpActionResult> Put(int id, Models.Shared.ServiceOfferingModel model)
+        {
```
Flemingjp, line 84: Same as above

`using ServiceOfferingModel = Biobanks.Web.Models.Shared.ServiceOfferingModel;`

src/Directory/Biobanks.Web/ApiControllers/AssociatedDataProcurementTimeFrameController.cs
```diff
@@ -0,0 +1,208 @@
...
+        }
+
+        [HttpPut]
+        [Route("{id}")]
+        public async Task<IHttpActionResult> Put(int id, Models.Shared.AssociatedDataProcurementTimeFrameModel model)
+        {
+            // Validate model
+            if (await _biobankReadService.ValidAssociatedDataProcurementTimeFrameDescriptionAsync(id, model.Description))
+            {
+                ModelState.AddModelError("AssociatedDataProcurementTimeFrame", "That Associated Data Procurement Time Frame already exists!");
```
beforan, line 93: This is an odd error message. This is a PUT on a resource id, which suggests it should be used for editing the resource with that id. In which case I would hope it does exist.

Or do we create at specific id's? (I don't think so; we also have a POST action for creating).

Is just the error message wrong? Or the actual validation method?

beforan: Not sure if this affects other controllers' PUT methods too

src/Directory/Biobanks.Web/ApiControllers/AssociatedDataProcurementTimeFrameController.cs
```diff
@@ -0,0 +1,208 @@
...
+            }
+
+            if (model.DisplayName == null)
+            {
+                ModelState.AddModelError("AssociatedDataProcurementTimeFrame", "The Display Name field is required.");
```
beforan, line 98: Why can't these be done by annotations on the model? Seems a little odd...
```
[Required]
[MinLength(10)]
```

src/Directory/Biobanks.Web/ApiControllers/AssociatedDataProcurementTimeFrameController.cs
```diff
@@ -0,0 +1,208 @@
...
+            {
+                success = true,
+                name = model.Description,
+            });
+
+        }
+
+        [HttpPut]
+        [Route("{id}")]
+        public async Task<IHttpActionResult> Put(int id, Models.Shared.AssociatedDataProcurementTimeFrameModel model)
```
beforan, line 88: prefer using statements to inline namespacing, unless clarifying an ambiguity between two or more identically named (but differently namespaced) types in use in the same file.

`using Models.Shared;`

especially since the type is used multiple times.

src/Directory/Biobanks.Web/Scripts/ADAC/adac-access-conditions.js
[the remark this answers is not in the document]
beforan: happy with making them consistent, didn't realise they didn't all use Edit ;)

src/Directory/Biobanks.Web/Scripts/ADAC/adac-access-conditions.js
```diff
@@ -128,11 +141,16 @@ $(function () {
         var triggerRow = diff.filter(row => row.node == edit.triggerRow.node())[0];
 
         //AJAX Update
-        $.post("EditAccessConditionAjax?sortOnly=true",
-            {
+        $.ajax({
+            url: $(triggerRow.node).data('resource-url') +
+                 "/Sort/" + $(triggerRow.node).data('access-condition-id'),
```
beforan, line 146: sorry, this route'll need changing as the controllers change

beforan (changes requested):

src/Directory/Biobanks.Web/ApiControllers/AssociatedDataProcurementTimeFrameController.cs
shakisparki: that is the case in this file (ambiguity). But I'll use the namespace on the other other one as it appears only once

src/Directory/Biobanks.Web/ApiControllers/AssociatedDataProcurementTimeFrameController.cs
beforan: oh it doesn't mean the record already exists (by id), it means the description/name/value is the same as another record's? that makes more sense, and can be left.

beforan (approved):

beforan (approved):

JosephLavagna (approved):

PR #53, opened 2020-11-23
Fix To Search and Facets

Flemingjp:
Fixes to search

- Fixed counting on facets; facets will now show how many of the relevant biobanks they apply to, rather than all biobanks on the directory

- Fixed an issue where non-nested facets were applied by their opposite logic (eg. Male facet showing only Female results)

- Added handling of edge case for Consent Restriction Facet - which is Opt-Out. Hence the behaviour is opposite to the other facets. This required manual calculation of the facet value's cardinality.

    -  Added handling for edge case, of this edge case. Where "No Restrictions" facet value reverts back to follow the default behaviour

### DIFF
--- a/src/Directory/Biobanks.Web/Views/Search/_Facets.cshtml
+++ b/src/Directory/Biobanks.Web/Views/Search/_Facets.cshtml
@@ -3,125 +3,125 @@
 @using Newtonsoft.Json
 @using Directory.Data.Constants
 @model FacetsModel
-@{
-    var currentGroup = "";
-}
 
 <div class="col-sm-4">
-    @if (!Model.Facets.IsNullOrEmpty())
-    {
-        @*var restrictionFacets = (IEnumerable<SearchFacetModel>)HttpContext.Current.Session["restrictionFacets"];*@
+	@if (!Model.Facets.IsNullOrEmpty())
+	{
+		foreach (var facetGroup in Model.Facets.GroupBy(x => x.GroupName))
+		{
 
-        foreach (var facet in Model.Facets)
-        {
-            if (currentGroup != facet.GroupName)
-            {
-                if (!string.IsNullOrEmpty(currentGroup))
-                {
-                    @* Close panel *@
-                    @Html.Raw("</div></div></div>")
-                }
+			// Facet Group Data
+			var firstFacet = facetGroup.First();
 
-                currentGroup = facet.GroupName;
+			var facetGroupNameId = $"FacetGroup{firstFacet.GroupNameWithoutSpaces}";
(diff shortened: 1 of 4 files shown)


src/Directory/Biobanks.Web/Views/Search/_Facets.cshtml
```diff
@@ -10,9 +10,9 @@
 <div class="col-sm-4">
     @if (!Model.Facets.IsNullOrEmpty())
     {
-        var restrictionFacets = (IEnumerable<SearchFacetModel>)HttpContext.Current.Session["restrictionFacets"];
+        @*var restrictionFacets = (IEnumerable<SearchFacetModel>)HttpContext.Current.Session["restrictionFacets"];*@
```
beforan, line 13: should this be uncommented? or removed?

src/Directory/Biobanks.Web/Views/Search/_Facets.cshtml
```diff
@@ -10,9 +10,9 @@
...
 
-        foreach (var facet in restrictionFacets)
+        foreach (var facet in Model.Facets)
```
beforan, line 15: Are we sure this doesn't break an intended feature?

Just someone bothered to get the restriction facets from the session (and presumably put them there?). Also, why only restriction facets?!

This seems very intentional, and presumably was done to save a current user's choices on refresh or something? Where/When is that key set in the Session?

src/Directory/Biobanks.Web/Views/Search/_Facets.cshtml
Flemingjp: Left it commented - like you mentioned it could be a breaking change. I have tested carefully, and checked the source and it appears to not be required. At this point I believe it to be left over debugging from when search was broken - but I could be wrong.

src/Directory/Biobanks.Web/Views/Search/_Facets.cshtml
beforan: It's set at line 69 in `SearchController`.
it's a recent-ish change if you look at the file history in the private repo.
If we don't need this session stuff, then we need to remove the setting from the controller too.

src/Directory/Biobanks.Web/Views/Search/_Facets.cshtml
beforan: Since we've removed the setting in the controller now, I think we should remove this line. it's always in git if we need it, but it's doing nothing without that session key being set.

beforan (approved):

blargly (dismissed):

beforan (approved):

blargly (approved):

PR #60, opened 2020-11-27
Removing Biobanks & BiobanksIds controllers from Directory and removing references in Analytics

bradthompson995:
The Biobanks & BiobanksIds api controllers are no longer neccesary so this PR removes them for the directory code and also removes the API calls from Analytics. Analytics has been modified to access the DB via local functions. Previous changes to master have done this for publications so for consistency a similar approach was used for doing the same to Analytics.


src/Analytics/AzureFunctions/Startup.cs
```diff
@@ -22,7 +21,7 @@ public override void Configure(IFunctionsHostBuilder builder)
                 .GetService<IConfiguration>();
 
             // Populate connection string with credentials
-            var sqlConnection = _configuration.GetConnectionString("analyticsdb_connection");
+            var sqlConnection = _configuration.GetConnectionString("sqldb-connection");
             var sqlUsername = _configuration.GetValue("sqldb-username", "");
             var sqlPassword = _configuration.GetValue("sqldb-password", "");
```
Flemingjp, line 27: Maybe just use an `sqldb-connection` string? I know this is copied from what I wrote in the Publications function, but the Directory now parses the connection string via its deployment pipeline. Changing to that way would keep the behaviour consistent across the repo.

src/Analytics/Analytics/Organisation.cs
```diff
@@ -0,0 +1,57 @@
+using System;
```
Flemingjp, line 1: Is this entity a temporary solution until we have a common `Entities` project?

Flemingjp (changes requested):
Everything look good - just a question and small change suggestion that might be useful while we're here

src/Analytics/Analytics/Organisation.cs
shakisparki: Yes it is

src/Analytics/AzureFunctions/Startup.cs
Flemingjp: @shakisparki  Don't forget to amend the pipelines

Flemingjp (approved):

src/Analytics/Services/GoogleAnalyticsReadService.cs
```diff
@@ -10,60 +10,55 @@
...
-        private readonly string DATE_FORMAT = "yyyy-MM-dd";
-
-        private readonly GoogleCredential credentials;
-        private readonly AnalyticsReportingService analytics;
+        private readonly string _viewId;
+        private readonly string _startDate; //default: "2016-01-01"; //specified in _dateFormat
+        private readonly string _dateFormat = "yyyy-MM-dd";
+        private readonly IConfiguration _config;
+
+        private AnalyticsDbContext _ctx;
```
beforan, line 34: pretty sure this should be `readonly`; it's dependency injected and assigned in the constructor only?

beforan (approved):

beforan (changes requested):
i misreviewed 😛 please resolve my comments 👍

beforan (approved):

Flemingjp (approved):

PR #62, opened 2020-11-30
Update README

Flemingjp:


README.md
```diff
@@ -15,15 +15,21 @@ The `src/` folder contains roughly project or app categorised folders, which als
 ## Directory structure:
 
 - `.azure/pipelines/` - Azure Pipelines configurations for conditionally building the different applications and libraries in the repo
-- `src/` - actual source code:
+- `elastic-search` - Configurational scripts to setup the required elastic search instance
```
beforan, line 18: I hate to be this petty, but can can we consistently end directories with `/`?

Partly so it's clear they're not files, but mainly because it's now inconsistent with the ones there from before.

beforan (changes requested):

beforan (approved):

shakisparki (approved):

PR #65, opened 2020-12-01
Register Sample Resource/Network Config 

JosephLavagna:
- Merged the Sample Resource and Network registration config pages
- Changed the label title to description in both cases
- Text fields now use markdown 
- Added two additional fields for the registration page titles


Flemingjp (approved):

src/Directory/Biobanks.Web/Views/Register/Biobank.cshtml
```diff
@@ -6,7 +6,7 @@
     ViewBag.Title = "Register a new " + App.Config[ConfigKey.SampleResourceName];
 }
 
-<h2>Register a new @App.Config[ConfigKey.SampleResourceName]</h2>
+<h2>@App.Config[ConfigKey.RegisterBiobankTitle]</h2>
```
beforan, line 9: Can you confirm which of the configurable fields are allowed to be markdown?

Is it the titles?

If not, that's fine.

If so, this is bad now, because you're putting arbitrary markdown rendered html inside an h2.

At best the CSS might be dumb, at worst you'll put invalid html tags inside an h2.

https://stackoverflow.com/questions/21228240/element-p-not-allowed-as-child-of-element-h2-in-this-context

JosephLavagna: @beforan The titles are just standard text inputs and the descriptions are markdown. That shouldn't be an issue right?

beforan: That sounds fine. Was just concerned the titles had been turned into markdown!

beforan (approved):

PR #66, opened 2020-12-02
SubmissionExpiryJob Azure Function

Flemingjp:
Moving ExpiryJob to an Azure Function


src/Api/SubmissionExpiryJob/.gitignore
```diff
@@ -0,0 +1,264 @@
+## Ignore Visual Studio temporary files, build results, and
```
Flemingjp, line 1: Will merge with root .gitignore

src/Api/SubmissionExpiryJob/Services/SubmissionService.cs
```diff
@@ -0,0 +1,50 @@
...
+                .Select(grp => new
+                {
+                    biobankid = grp.Key,
+                    lastUpdatedSubmission = grp.OrderByDescending(x => x.StatusChangeTimestamp)
+                        .FirstOrDefault(lastUpdate =>
+                            DateTime.Now.Subtract(TimeSpan.FromDays(_settings.ExpiryDays)) > lastUpdate.StatusChangeTimestamp)
+                })
+                .Select(s => s.biobankid)
+                .ToListAsync();
+        }
```
Flemingjp, line 37: Am I right in thinking the first Select statement does nothing? Is there a missing statement after `Where(x => x.lastUpdatedSubmission != null)`

src/Api/SubmissionExpiryJob/Services/SubmissionService.cs
```diff
@@ -0,0 +1,59 @@
...
+
+            // Mark Submissions As Rejected
+            foreach (var submission in _db.Submissions.Where(s => s.BiobankId == organisationId))
+            {
+                submission.StatusChangeTimestamp = DateTime.Now;
+                submission.Status = await _db.Statuses.FirstOrDefaultAsync(s => s.Value == Statuses.Rejected);
+            }
+
+            await _db.SaveChangesAsync();
+        }
```
Flemingjp, line 57: This behavoir is coped from the `/reject` endpoint. It currently will find biobanks with a single expired submission, and proceed to delete all submission data. Is this desired?

beforan: what does the current expiry webjob do?

Flemingjp: Just marks the submission as rejected. Leaving the submission data in the table(s)

beforan: but no, it's not desirable to expire (or reject) ALL submissions for that bb, only open ones.

src/Api/SubmissionExpiryJob/Startup.cs
```diff
@@ -0,0 +1,47 @@
...
+
+            if (!string.IsNullOrEmpty(appInsightsKey))
+            {
+                builder.Services.AddApplicationInsightsTelemetry(appInsightsKey);
+            }
+
+            // Settings From Configuration
+            builder.Services.Configure<SettingsModel>(options =>
+            {
+                options.ExpiryDays = _configuration.GetValue<int>("expiryDays", 0);
```
beforan, line 33: is `0` a desirable default?

It seems pretty harsh. We should either default to something generous, or perhaps allow expiry to be off, and use that by default.

Also, need to document configuration options and default behaviours

src/Api/Biobanks.Apis.sln
```diff
@@ -19,6 +19,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "IdentityProvider", "Identit
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SubmissionJob", "SubmissionJob\SubmissionJob.csproj", "{3027943B-B5AC-46A0-8663-E677DE81C0AC}"
```
beforan, line 20: This is the old webjob project? We're not getting rid of it?

src/Api/Biobanks.Apis.sln
```diff
@@ -19,6 +19,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "IdentityProvider", "Identit
...
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SubmissionExpiryJob", "SubmissionExpiryJob\SubmissionExpiryJob.csproj", "{543B1D91-97B7-498D-8F84-4917EF3F1C0A}"
```
beforan, line 22: wouldn't both the functions (expiry and processing) live in the same functions project? but just be different functions with different triggers?

Honestly not sure what the best practice is, but I don't believe they need to be separate projects like webjobs did, and I'm nervous of the effect on Azure resources of them being separate.

src/Api/SubmissionExpiryJob/Services/SubmissionService.cs
Flemingjp: I agree - the problem currently is that for a Staged Submission, it has no reference to the Submission it is part of. The Staged Submission only reference the Biobank ID and not the Submission ID. They can be loosely linked by having the same timestamp - but that doesn't guarantee it to be unique in all cases. 

This same problem seems to crop up in the `reject` endpoint, where it flat out delete *all* staged submissions for a given biobank when a single submission is rejected.

Am I misunderstanding the current submission process - can biobanks only have multiple open submissions at a time?

src/Api/SubmissionExpiryJob/Startup.cs
Flemingjp: I agree for a generous default, say `30 days`. To disable the expiry function - just stop the azure function?

src/Api/Biobanks.Apis.sln
Flemingjp: Good point - Ive have them as two separate functions due how the Analytics and Publicatiosn were done. However here they have a shared context. 

If they're the same function, its be a singular function on Azure - so can be done with one function. What would be best for project structure? One Azure Functions project and then two class libraries?

It will simplify things in one way - but complex them more in another - such as Dependency Injection

src/Api/Biobanks.Apis.sln
Flemingjp: Was going to remove the two old webjobs once the new azure functions were merged. Try keeping this PR clean of a few hundred deleted files

src/Api/SubmissionExpiryJob/Services/SubmissionService.cs
Flemingjp: So this is what I believe is the correct, and implemented, behaviour.

- If the latest submission from a Biobank is older than the expiry threshold, then all currently open submissions (that have therefore also expired) are flagged as expired
- All data for that biobank is deleted, as the data is older than the expiry threshold

Flemingjp:
Succeeded by #73

PR #67, opened 2020-12-02
Remove Submission API ref data endpoints

JosephLavagna:
Deleted all controller classes contained within the refdata folder which were composed of a number of CRUD endpoints.


Flemingjp (dismissed):

Flemingjp (approved):

Flemingjp:
Question before merge - Should the Reference Data services also go? Most of them are only used by the API Controllers that are being removed.

PR #70, opened 2020-12-07
Publications build pipeline

shakisparki:


src/Publications/Publications/PublicationDbContext.cs
```diff
@@ -20,18 +20,4 @@ public class PublicationDbContext : DbContext
...
-        public PublicationDbContext CreateDbContext(string[] args)
-        {
-            var options = new DbContextOptionsBuilder<PublicationDbContext>();
-
-            // Connection string passed from environment variable as EF doesn't (yet) support supplying it as a CLI parameter
-            options.UseSqlServer(Environment.GetEnvironmentVariable("sqldb_connection"), options => options.EnableRetryOnFailure());
-
-            return new PublicationDbContext(options.Options);
-        }
-    }
```
Flemingjp, line 36: What is the reason for removing this part - I know it solved a prior issue with migrations. Is this resolved now?

shakisparki: I believe that was to run migrations from the pipeline while passing the connection string as an env var.
Since we aren't running migrations from publications anymore (entities are maintained from the directory), then we shouldn't need this

Flemingjp: Yepp that sounds right - does that include the `Publication` entity?

shakisparki: Yeah it does

Flemingjp (approved):

beforan (approved):

PR #71, opened 2020-12-07
Upgrade submission API to .NET 5, Remove Join Entities, New Data and LegacyData projs

JosephLavagna:
- Upgrade the submission API to .NET 5 
- Removed the material type/group join entity
- Added Data Project which stores the master DbContext - .NET Core 5 and EF Core 5
- Added LegacyData Project which stores another version of DbContext - .NET Core 3.1 and EF Core 2
- Removed old migrations
-  Created new migrations in the Data project
- API now uses the DbContext in the Data project
- Webjobs/Functions need to be updated to use the LegacyData project once merged


src/Api/Common/Data/Entities/ReferenceData/MaterialType.cs
```diff
@@ -23,6 +22,6 @@ public class MaterialType
         /// <summary>
         /// Join entities for Material Type Groups that this type is a member of.
         /// </summary>
```
Flemingjp, line 24: Maybe worth updating the summary text for these functions - as it will become contradictory

src/Api/SubmissionApi/Services/CommitService.cs
```diff
@@ -34,15 +34,15 @@ public async Task CommitStagedData(bool replace, int organisationId)
...
+                await _db.Database.ExecuteSqlRawAsync(SqlConsts.DeleteAllLiveTreatments(organisationId));
             }
 
             // merge staged entities into corresponding live tables
-            await _db.Database.ExecuteSqlCommandAsync(SqlConsts.MergeStagedDiagnosesIntoLive(organisationId));
-            await _db.Database.ExecuteSqlCommandAsync(SqlConsts.MergeStagedSamplesIntoLive(organisationId));
-            await _db.Database.ExecuteSqlCommandAsync(SqlConsts.MergeStagedTreatmentsIntoLive(organisationId));
+            await _db.Database.ExecuteSqlRawAsync(SqlConsts.MergeStagedDiagnosesIntoLive(organisationId));
+            await _db.Database.ExecuteSqlRawAsync(SqlConsts.MergeStagedSamplesIntoLive(organisationId));
+            await _db.Database.ExecuteSqlRawAsync(SqlConsts.MergeStagedTreatmentsIntoLive(organisationId));
```
Flemingjp, line 46: What is the reason for the change in function - is it due to the update to .NET 5?

Flemingjp (dismissed):

Flemingjp (changes requested):
Just the above comments

src/Api/SubmissionApi/Services/CommitService.cs
JosephLavagna: Yep this is because of package update - the ExecuteSqlCommandAsync method was no longer available.

src/Api/Common/Common.csproj
```diff
@@ -17,4 +17,8 @@
     <PackageReference Include="Z.EntityFramework.Plus.EFCore" Version="1.8.10" />
   </ItemGroup>
 
+  <ItemGroup>
```
beforan, line 20: this ItemGroup can go

Flemingjp (approved):

src/Api/SubmissionApi/SubmissionApi.csproj
```diff
@@ -1,12 +1,13 @@
...
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework>netcoreapp5.0</TargetFramework>
     <DocumentationFile>swagger.xml</DocumentationFile>
-    <VersionPrefix>2.1.2</VersionPrefix>
+    <VersionPrefix>3.0</VersionPrefix>
     <RootNamespace>Biobanks.SubmissionApi</RootNamespace>
     <AssemblyName>SubmissionApi</AssemblyName>
     <UserSecretsId>216e442f-ab1c-4cb2-8e0e-ed462d282d2b</UserSecretsId>
+	 <AddRazorSupportForMvc>true</AddRazorSupportForMvc>
```
beforan, line 10: why? We aren't using any razor views, it's an API.

src/Api/SubmissionApi/SubmissionApi.csproj
```diff
@@ -1,12 +1,13 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework>netcoreapp5.0</TargetFramework>
     <DocumentationFile>swagger.xml</DocumentationFile>
-    <VersionPrefix>2.1.2</VersionPrefix>
+    <VersionPrefix>3.0</VersionPrefix>
```
beforan, line 6: not sure who updated this to 2.1.2 but this is part of the version of the app (SubmissionApi), not anything to do with .NET or anything.

I'd prefer to use the plain `Version` property, like DECSYS does:
`<Version>2.0.0-beta.7</Version>` for example.

As for what version this is... I guess we can call it `3.0.0-alpha.1` so as not not be confusing :\

beforan: (since 2.1.x has already been used even though 2.0 never actually got released. that's the confusion)

src/Api/SubmissionApi/SubmissionApi.csproj
```diff
@@ -20,23 +21,26 @@
...
 
   <ItemGroup>
-    <PackageReference Include="Autofac.Extensions.DependencyInjection" Version="4.3.0" />
+    <PackageReference Include="Autofac.Extensions.DependencyInjection" Version="7.1.0" />
     <PackageReference Include="AutoMapper" Version="7.0.1" />
     <PackageReference Include="AutoMapper.Extensions.Microsoft.DependencyInjection" Version="5.0.1" />
     <PackageReference Include="clacks.overhead" Version="1.0.0" />
-    <PackageReference Include="Hangfire" Version="1.6.20" />
-    <PackageReference Include="Hangfire.SqlServer" Version="1.6.20" />
+    <PackageReference Include="Hangfire" Version="1.7.18" />
```
beforan, line 28: What do we use Hangfire for in this project?!

src/Data/SubmissionsDbContext.cs
```diff
@@ -0,0 +1,117 @@
...
+                    x.Barcode,
+                    x.CollectionName
+                }).IsUnique();
+
+        }
+
+        public SubmissionsDbContext(DbContextOptions options) : base(options) { }
+
+        /*  Required for DesignTime creation of the context.*/
+        public class SubmissionsDbContextFactory : IDesignTimeDbContextFactory<SubmissionsDbContext>
```
beforan, line 103: don't believe this is needed, but I think you're testing the command line connection string stuff for migrations?

src/LegacyData/LegacyData.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
beforan, line 4: if we're still using the `2.2.x` packages, we should still target `netcoreapp2.2`.

everything will go up to `3.1` when we do the Azure Functions merge later

src/LegacyData/SubmissionsDbContext.cs
```diff
@@ -94,15 +95,10 @@ protected override void OnModelCreating(ModelBuilder model)
                     x.CollectionName
                 }).IsUnique();
 
-            //Join entity keys
-            model.Entity<MaterialTypeMaterialTypeGroup>()
```
beforan, line 98: do the webjobs use materialtype/materialtypegroups? I assume so. They will still need the join entity, since the automatic one is an ef core 5 feature...

But the join entity itself and the two entities that use it (MT and MTG) can move local to this project instead of Common. Think of it like an ef core 3.1 (or 2,2) specific override.

beforan (changes requested):

src/Api/SubmissionApi/SubmissionApi.csproj
Flemingjp: It appears to be used in the `/commit` and `/reject` endpoint to queue the actual business logic as an async background task.

src/LegacyData/LegacyData.csproj
Flemingjp: Would it be ideal for 3.1 now, if the Azure function is going to merge right after?

src/Api/SubmissionApi/SubmissionApi.csproj
beforan: oh ffs. then why do we even use webjobs/az functions? why don't we just use hangfire for everything?! hmm, that violates the original design of this app somewhat. if those operations need to be background, they should have their own functions like submissionjob... We deliberately opted not to use hangfire so background tasks didn't run off the same process / server.

beforan: Sorry, I'm angry at the past.

beforan: Hangfire here may cause a problem, because the Directory today *does* use hangfire intentionally, but they may get in each other's way using the same tables... I think we need more functions. Which also means we need more queueing in az storage accounts. Also, what does the REST API give back for those endpoints? it should be 202 Accepted if the request is received but the response returns before it is actioned. Like Submit.

Flemingjp: Looking at the `RejectService` the logic is identical to the logic in the Azure Expiry function - with the exception of `Status.Rejected` vs `Status.Expired`

Flemingjp: The actions currently return `NoContent()`

src/Data/SubmissionsDbContext.cs
beforan: it is still needed; we're working out the details ;)

src/Api/SubmissionApi/SubmissionApi.csproj
JosephLavagna: @beforan @Flemingjp what's the verdict then?

Flemingjp: I'm assuming its going to be another Azure function - looking at the code it is identical to the current expiry function, except it marks the submission as rejected. My question is do we just duplicate the Azure expiry function and change it to a reject function - or have one function that does both with a flag (reject or expire).

Personally, I think (maybe) duplicate it, in case the behaviour diverges in the future, and keeps the purpose of the functions clear.

beforan: you can (probably should?) factor out the actual work piece into a service method, then have separate expiry and reject functions (in part because they are triggered differently, too) that today call the same service method, but in future it would be easy for them not to.

but yeah, expiry is timed (nightly) and reject is a queue trigger presumably, or an http trigger if you really want?

beforan: Agree that for this PR we can leave hangfire, having gained all this knowledge and committed to implement it in the functions PR

src/LegacyData/LegacyData.csproj
beforan: yeah I think so.

beforan: but in that case the packages should get upgraded 3.1.x ;)

src/Api/SubmissionApi/SubmissionApi.csproj
```diff
@@ -1,9 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework>netcoreapp5.0</TargetFramework>
```
beforan, line 4: .NET 5's target moniker is `net5.0` just to be different ;)

beforan (changes requested):

src/LegacyData/Entities/MaterialType.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// <summary>
+        /// Value for the Material Type (i.e. friendly name)
+        /// </summary>
+        [Required]
+        public string Value { get; set; }
+
+        /// <summary>
+        /// Many to Many Relationship with MaterialTypeGroup
+        /// </summary>
+        public ICollection<MaterialTypeGroup> MaterialTypeGroups { get; set; }
```
beforan, line 25: Doesn't this one need to go via the Join Entity?

src/LegacyData/Entities/MaterialTypeGroup.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// <summary>
+        /// The friendly name of the group.
+        /// </summary>
+        [Required]
+        public string Value { get; set; }
+
+        /// <summary>
+        /// Join entities for MaterialTypes in the group
+        /// </summary>
+        public ICollection<MaterialType> MaterialTypes { get; set; }
```
beforan, line 27: Join Entity?

beforan (changes requested):

Flemingjp (approved):
Approved - just need fixes for the current comments

Flemingjp (approved):

beforan (approved):

beforan:
merging this with api builds failing under the assumption that #73 will fix it

PR #72, opened 2020-12-07
delete biobank registration request when biobank is deleted

shakisparki:
Allow biobank of the same name to be re-created after deletion
-  by deleting registration request as part of the process of deletion of all associated data


Flemingjp (approved):

src/Directory/Services/BiobankReadService.cs
```diff
@@ -614,6 +614,17 @@ public async Task<int> GetSuspendedCapabilityCountAsync()
                     })
                 .ToDictionary(x => x.id, x => x.description);
 
+        public async Task<OrganisationRegisterRequest> GetBiobankRegisterRequestByOrganisationNameAsync(string name)
+        {
+            //We consider declined requests to not exist
```
beforan, line 619: this comment is a lie. you've removed the `DeclinedDate == null` criteria from the filter. Is that desirable? (I think so, probably). If so, remove this lying comment :P

beforan (changes requested):

beforan (approved):

Flemingjp (approved):

PR #73, opened 2020-12-08
Submission API Azure Functions

Flemingjp:
A single project, azure function for running two background processes for the API
- Staging submitted data from the API
- Expiring submissions of biobanks
- Removal of old Submission and Expiry WebJob Projects

PR #74, opened 2020-12-08
API Submission DateTime Validation

Flemingjp:
### Changes
- Removed submission validation of DateTime being in the future - this validation should be done at the Staging Azure Function rather that at the submit controller

- Changed all submission payload models to use a nullable DateTime quantity, such that missing DateTime values are parsed as null, rather than a default value. 

- Tested against payloads with singular and multiple entries

### New Behaviour

- If at least one submission entry has an illegally formatted DateTime (eg "2012-AB-40") then the whole request is rejected with **HTTP 400 Bad Request**, with information on which field is illegally formatted.

- If at least one submission entry has a missing DateTime field, then the whole request is rejected with a **HTTP 400 Bad Request** with information on which entry has the missing fields.

- If all submission have legally formatted DateTime fields, then the submission is accepted with a **HTTP 202 Accepted**, regardless if the DateTime is past, present or future.


src/Api/Common/Models/DiagnosisIdModel.cs
```diff
@@ -10,7 +10,7 @@ public class DiagnosisIdModel : BaseIdModel
         /// The date of the diagnosis.
         /// </summary>
         [Required]
-        public DateTime DateDiagnosed { get; set; }
+        public DateTime? DateDiagnosed { get; set; }
```
Flemingjp, line 13: Some models were already were marked as `[Required]`  and now are nullable type. Not sure the exact interaction here, but the generated Swagger documentation does not flag the quantity as either required or nullable. Thoughts @beforan

src/Api/SubmissionApi/Models/TreatmentSubmissionModel.cs
```diff
@@ -10,6 +10,7 @@ public class TreatmentSubmissionModel : TreatmentIdModel
         /// <summary>
         /// The anatomical location where the treatment was performed.
         /// </summary>
+        [Required]
         public string TreatmentLocation { get; set; }
```
Flemingjp, line 14: Added `[Required]` to `TreatmentLocation` as the Swagger Documentation didn't flag this field as required, even though it is via the Submit controller validation.

Flemingjp:
Waiting on #73

blargly (approved):

src/Api/Common/Models/DiagnosisIdModel.cs
beforan: We resolved this offline. `[Required]` nullable here provides the intended behaviour:

- nullable stops the ASP.NET ModelBinder initialising the property with the default value if it's missing from a request payload
- `[Required]` ensures we validate the presence of the property in the model; a `null` value will fail validation.

src/Api/SubmissionApi/Controllers/SubmitController.cs
```diff
@@ -153,7 +153,7 @@ public async Task<IActionResult> Post(int biobankId, [FromBody] SubmissionModel
                             return await CancelSubmissionAndReturnBadRequest(sampleModel, submission.Id, "Invalid StorageTemperature value.");
                         else if (sampleModel.AgeAtDonation == null && sampleModel.YearOfBirth == null)
                             return await CancelSubmissionAndReturnBadRequest(sampleModel, submission.Id, "At least one of AgeAtDonation or YearOfBirth must be provided.");
-                        else if (sampleModel.DateCreated == default(DateTime) || sampleModel.DateCreated > DateTime.Now)
+                        else if (sampleModel.DateCreated == null)
```
beforan, line 156: these null checks should be redundant now as `[Required]` should catch the null values in the [ModelState check](https://github.com/biobankinguk/biobankinguk/pull/74/files#diff-e3a182992df947001a3d991380157b86671bf607d9216f6499b347515447e291R68)

Can we remove them if they're not needed?

src/Api/SubmissionApi/Models/TreatmentSubmissionModel.cs
beforan: Similarly to the dates, `[Required]` should make it caught by `ModelState.IsValid`. Is there a reason Required was left off in order to catch it in the later validation? (Does the later validation do something different?)

beforan (changes requested):

src/Api/SubmissionApi/Models/TreatmentSubmissionModel.cs
beforan: I mean, that was a question :P If the validation doesn't do anyting special (doesn't look like it does anything different to the dates) then `[Required]` is fine and we can remove the later validation, as `ModelState.IsValid` will do it for us, earlier, which I think is preferable...

beforan (approved):

beforan (approved):

JosephLavagna (approved):

PR #79, opened 2020-12-14
Added API Submissions Azure Function Build Pipeline

Flemingjp:


.azure/pipelines/build.api.submissions.yml
```diff
@@ -0,0 +1,40 @@
...
+# https://docs.microsoft.com/en-us/azure/devops/pipelines/languages/dotnet-core 
+
+trigger:
+  batch: true
+  branches:
+    include:
+    - "*"
+  paths:
+    include:
+    - "src/Api/SubmissionAzureFunction/*"
```
beforan, line 12: pretty sure this should also depend on `LegacyData` and possibly `Common` and possibly an `Entities` project? Basically any project references should also be here, as changes to their code could break this app.

beforan (changes requested):

beforan (approved):

blargly (approved):

PR #80, opened 2020-12-14
Claims replacement

blargly:
This replaces the filter OrganisationsRoleAttribute. This called on the database 4 times every time a controller method was run, in order to populate the viewbag with biobank, network, biobank requests and network requests ID values and names.
However, this same information is queried in the custom claims manager (using the same methods), so these claims have now been expanded to include the name for the 4 types. 

Any bits of the code which previously used the viewbag for biobank Ids now looks up claims. 

Furthermore, existing bits of code that use claims need to deserialise first due to the addition of a name.

Currently, a draft as I have some thoughts on restructuring this to centralise all claims access effected by this change to use the ApplicationUserPrincipal


src/Directory/Identity/ApplicationUserPrincipal.cs
```diff
@@ -12,8 +13,9 @@ public class ApplicationUserPrincipal : ClaimsPrincipal
 
         //Add custom claims...
         public string Name => FindFirst(CustomClaimType.FullName).Value; //ClaimTypes.Name is used for username, which is acquirable directly from the identity instead
-        
-        public IEnumerable<string> BiobankIds => FindAll(CustomClaimType.BiobankId).Select(x => x.Value);
-        public IEnumerable<string> NetworkIds => FindAll(CustomClaimType.NetworkId).Select(x => x.Value);
+
+        public IEnumerable<string> BiobankIds => FindAll(CustomClaimType.BiobankId).Select(x => JsonConvert.DeserializeObject<KeyValuePair<string, string>>(x.Value).Key);
```
blargly, line 17: Having made yet another Deserialise call I wonder if perhaps BiobankIds should be replaced with a IENumerable<KeyValuePair<int, string> Biobanks (and likewise new claims for requests) . Anywhere we then try and filter currentuser to get the correct claim types, we instead just get the user property managed here.

Flemingjp: Is there a difference between `IEnumerable<KeyValuePair<int, string>>` vs  `Dictionary<int, string>`? Is it possible to deserialize into a `Dictionary`.

src/Directory/Biobanks.Web/Controllers/AccountController.cs
```diff
@@ -188,23 +188,32 @@ public async Task<ActionResult> LoginRedirect(string returnUrl = null)
...
             {
-                List<KeyValuePair<int, string>> biobankList = JsonConvert.DeserializeObject<List<KeyValuePair<int, string>>>(ViewBag.UserBiobanks);
-                var biobank = biobankList.FirstOrDefault();
-                Session[SessionKeys.ActiveOrganisationType] = ActiveOrganisationType.Biobank;
-                Session[SessionKeys.ActiveOrganisationId] = biobank.Key;
-                Session[SessionKeys.ActiveOrganisationName] = biobank.Value;
-                return RedirectToAction("Collections", "Biobank");
+
+                var biobank = JsonConvert.DeserializeObject<KeyValuePair<int, string>>(CurrentUser.Claims
+                    .Where(x => x.Type == CustomClaimType.BiobankId).ToList().FirstOrDefault()?.Value);
```
beforan, line 193: have you reused the `BiobankId` ClaimType then? Waht about the auth places where it was used previously and expects only the integer id?

src/Directory/Biobanks.Web/Controllers/AccountController.cs
```diff
@@ -188,23 +188,32 @@ public async Task<ActionResult> LoginRedirect(string returnUrl = null)
...
+
+                if (!biobank.Equals(default(KeyValuePair<int, string>)))
```
beforan, line 195: I feel like it would be neater to try and get the claim first, then null check that, instead of default checking the JSON conversion.

src/Directory/Biobanks.Web/Controllers/AccountController.cs
beforan: Also you can simplify the Claims bit to: `CurrentUser.Claims.FirstOrDefault(x => condition)`

No need to `ToList()` early (I don't believe?), and `FirstOrDefault()` accepts a filter so you don't need `Where()`.

src/Directory/Biobanks.Web/Filters/AuthorizeAttribute.cs
```diff
@@ -10,7 +10,6 @@
 
 namespace Biobanks
 {
-    // TODO: what is this filter? is it used anywhere?! UserAuthorize seems to be used instead...
```
beforan, line 13: thanks

src/Directory/Biobanks.Web/Controllers/AccountController.cs
```diff
@@ -188,23 +188,32 @@ public async Task<ActionResult> LoginRedirect(string returnUrl = null)
...
+                {
+                    Session[SessionKeys.ActiveOrganisationType] = ActiveOrganisationType.Biobank;
+                    Session[SessionKeys.ActiveOrganisationId] = biobank.Key;
+                    Session[SessionKeys.ActiveOrganisationName] = biobank.Value;
+                    return RedirectToAction("Collections", "Biobank");
+                }
             }
                 
             // Single network only
```
beforan, line 204: I think this comment is a lie? the whole point is this works for multiple (but defaults to the first), isn't it? Or is this really only the case where they only admin one (I don't see any checks for that)

src/Directory/Biobanks.Web/Views/Header/_NavItemBiobanks.cshtml
```diff
@@ -1,32 +1,28 @@
 ﻿@using Newtonsoft.Json
+@using Directory.Identity.Constants;
 @model Biobanks.Web.Models.Header.NavItemModel
 
 @if (CurrentUser.Identity.IsAuthenticated)
 {
-	if (ViewBag.UserBiobankRequests != null)
+	var biobankRequestClaims = CurrentUser.Claims.Where(x => x.Type == CustomClaimType.BiobankRequestId).ToList();
```
beforan, line 7: Nothing gained from `ToList()`? It's only used for the `foreach` and that can do it over an `IEnumerable` quite happily.

src/Directory/Identity/ApplicationUserPrincipal.cs
beforan: @blargly since these properties already exist, yeah we should probably either replace these (and update whatever uses them) or add properties that are similar but contain the id and name to the principal, so simplify the controller and view code (that you've already written :P) yeah.

@Flemingjp
- Is there a difference? yes.
- Can it be a dictionary? Probably, but I dunno is JsonConvert would do that for us or if we'd have to write code to get it to taht point, in which case it's probably not worth it. Considering the use case (e.g. nav menus) We're actually more interested in looping over them all and using both the key and the value, than we are in looking up the value for a given key. So really an array of key value pairs is closer to what we want, as well as closer to what we are dealing with (which is an array of claims containing a serialised key value pair). Blah.
- There are not duplicates of the same key. Or at least there shouldn't be

src/Directory/Identity/Constants/CustomClaimType.cs
```diff
@@ -5,5 +5,7 @@ public static class CustomClaimType
         public const string FullName = "FullName";
         public const string BiobankId = "BiobankId";
         public const string NetworkId = "NetworkId";
+        public const string BiobankRequestId = "BiobankRequestId";
```
beforan, line 8: I think all these claim types need `Id` taking off the name, if they store more than just an Id.

beforan (changes requested):

src/Directory/Biobanks.Web/Controllers/AccountController.cs
beforan: This might get simpler anyway if the user principal provides the values rather  than deserializing here.

blargly: Yeah, this is now simplified as the deserialising happens centrally now.
As for checks, I went with using default as KeyValuePair is a struct, and will be initialised with default values for its key and value.

src/Directory/Biobanks.Web/Controllers/AccountController.cs
blargly: Not my comment, but yes it's not right - updated with // Network admin only as its actually based on if you administer networks only.

src/Directory/Biobanks.Web/Views/Header/_NavItemBiobanks.cshtml
blargly: Removed (and changed anyway due to the overall centralisation of claims logic).

src/Directory/Identity/Constants/CustomClaimType.cs
blargly: Agreed and done

Flemingjp (dismissed):

src/Directory/Biobanks.Web/Controllers/AccountController.cs
```diff
@@ -179,32 +179,39 @@ public async Task<ActionResult> LoginRedirect(string returnUrl = null)
...
+            // Network admin only
             if (CurrentUser.IsInRole(Role.NetworkAdmin.ToString()))
             {
-                List<KeyValuePair<int, string>> networkList = JsonConvert.DeserializeObject<List<KeyValuePair<int, string>>>(ViewBag.UserNetworks);
-                var network = networkList.FirstOrDefault();
-                Session[SessionKeys.ActiveOrganisationType] = ActiveOrganisationType.Network;
-                Session[SessionKeys.ActiveOrganisationId] = network.Key;
-                Session[SessionKeys.ActiveOrganisationName] = network.Value;
-                return RedirectToAction("Biobanks", "Network");
+                var network = JsonConvert.DeserializeObject<KeyValuePair<int, string>>(CurrentUser.Claims
```
beforan, line 205: shouldn't this use CurrentUser.Networks similar to the Biobanks case above?

src/Directory/Biobanks.Web/Controllers/ProfileController.cs
```diff
@@ -39,7 +40,7 @@ public async Task<ActionResult> Biobank(string id)
             if (bb.IsSuspended)
             {
                 //Allow ADAC or this Biobank's admins to view the profile
-                if (CurrentUser.BiobankIds.Contains(bb.OrganisationId.ToString()) && User.IsInRole(Role.BiobankAdmin.ToString()) ||
+                if (CurrentUser.Biobanks.FirstOrDefault(x => x.Key == Convert.ToInt32(bb.OrganisationId)).Key != 0 && User.IsInRole(Role.BiobankAdmin.ToString()) ||
```
beforan, line 43: it's not clear what's going on here.

I _think_ it's because of the `KeyValuePair` comparison instead of the old `string`, `FirstOrDefault` won't return `null` it'll be `default(KeyValuePair)` which you noted is a struct so the Key will have `default(int)` which is `0`...

- Is `FirstOrDefault` right? should it be `SingleOrDefault` (can keys be duplicated)?
- should it use `default(int)` instead of `0`?
- Why does `OrganisationId` need converting to an `int`? Isn't it already one?
- Is it worth a comment to indicate what's going on?
- Alternatively can we rewrite it in terms of a key comparison (which is what it's doing nested inside a KVP, and closer to the previous code), which would probably make it clearer (but use slightly more code)?
e.g.
```csharp
var keys = CurrentUser.Biobanks.Select(x => x.Key);
if(keys.Contains(bb.OrganisationId)) {
  // ...
}
```

src/Directory/Biobanks.Web/Filters/AuthoriseToAdministerCapability.cs
```diff
@@ -19,9 +19,9 @@ public override void OnAuthorization(AuthorizationContext filterContext)
             var session = filterContext.RequestContext.HttpContext.Session;
             var activeOrganisationId = session[SessionKeys.ActiveOrganisationId];
 
-            if (activeOrganisationId != null 
-                && Convert.ToInt32(session[SessionKeys.ActiveOrganisationType]) == (int) ActiveOrganisationType.Biobank
-                && filterContext.RequestContext.HttpContext.User.ToApplicationUserPrincipal().BiobankIds.Contains(activeOrganisationId.ToString()))
+            if (activeOrganisationId != null
+                && Convert.ToInt32(session[SessionKeys.ActiveOrganisationType]) == (int)ActiveOrganisationType.Biobank
+                && filterContext.RequestContext.HttpContext.User.ToApplicationUserPrincipal().Biobanks.FirstOrDefault(x => x.Key == Convert.ToInt32(activeOrganisationId)).Key != 0)
```
beforan, line 24: As above, this is really opaque and requires brain engagement each time!

maybe a helper method (possibly a `List<KeyValuePair<T1,T2>` Extension?) of `ContainsKey(T1 key)` would be good!

src/Directory/Biobanks.Web/Filters/AuthoriseToAdministerCollection.cs
```diff
@@ -21,8 +21,8 @@ public override void OnAuthorization(AuthorizationContext filterContext)
 
             if (activeOrganisationId != null
                 && Convert.ToInt32(session[SessionKeys.ActiveOrganisationType]) == (int) ActiveOrganisationType.Biobank
-                && filterContext.RequestContext.HttpContext.User.ToApplicationUserPrincipal().BiobankIds
-                    .Contains(activeOrganisationId.ToString()))
+                && filterContext.RequestContext.HttpContext.User.ToApplicationUserPrincipal()
+                .Biobanks.FirstOrDefault(x => x.Key == Convert.ToInt32(activeOrganisationId)).Key != 0)
```
beforan, line 25: same again

src/Directory/Biobanks.Web/Filters/AuthoriseToAdministerSampleSet.cs
```diff
@@ -21,8 +21,8 @@ public override void OnAuthorization(AuthorizationContext filterContext)
 
             if (activeOrganisationId != null
                 && Convert.ToInt32(session[SessionKeys.ActiveOrganisationType]) == (int) ActiveOrganisationType.Biobank
-                && filterContext.RequestContext.HttpContext.User.ToApplicationUserPrincipal().BiobankIds
-                    .Contains(activeOrganisationId.ToString()))
+                && filterContext.RequestContext.HttpContext.User.ToApplicationUserPrincipal()
+                .Biobanks.FirstOrDefault(x => x.Key == Convert.ToInt32(activeOrganisationId)).Key != 0)
```
beforan, line 25: again

src/Directory/Biobanks.Web/Filters/AuthorizeAttribute.cs
```diff
@@ -65,7 +64,7 @@ protected override bool AuthorizeCore(HttpContextBase httpContext)
                     var biobankId = Convert.ToInt32(activeOrganisationId);
 
                     // If they don't have a claim on this biobank, return
-                    if (!currentUser.BiobankIds.Contains(activeOrganisationId.ToString()))
+                    if (currentUser.Biobanks.FirstOrDefault(x => x.Key == Convert.ToInt32(activeOrganisationId)).Key == 0)
```
beforan, line 67: revenge of the son of the same thing again, part II

src/Directory/Biobanks.Web/Filters/AuthorizeAttribute.cs
```diff
@@ -92,7 +91,7 @@ await BiobankReadService.GetBiobankByIdAsync(biobankId))
                     var networkId = Convert.ToInt32(activeOrganisationId);
 
                     // If they don't have a claim on this biobank, return
-                    if (!currentUser.NetworkIds.Contains(networkId.ToString()))
+                    if (currentUser.Networks.FirstOrDefault(x => x.Key == networkId).Key == 0)
```
beforan, line 94: and same again, though without the int conversion?

src/Directory/Identity/ApplicationUserPrincipal.cs
```diff
@@ -12,8 +13,9 @@ public class ApplicationUserPrincipal : ClaimsPrincipal
 
         //Add custom claims...
         public string Name => FindFirst(CustomClaimType.FullName).Value; //ClaimTypes.Name is used for username, which is acquirable directly from the identity instead
-        
-        public IEnumerable<string> BiobankIds => FindAll(CustomClaimType.BiobankId).Select(x => x.Value);
-        public IEnumerable<string> NetworkIds => FindAll(CustomClaimType.NetworkId).Select(x => x.Value);
+        public IEnumerable<KeyValuePair<int, string>> Biobanks => FindAll(CustomClaimType.Biobank).Select(x => JsonConvert.DeserializeObject<KeyValuePair<int, string>>(x.Value));
```
beforan, line 16: Should we make these `List`? That would guarantee they are all enumerated (wastefully if they re never used), but if we already know they're going to be enumerated later (I think we do), it's better to do it just once now. That also makes the suggested ContainsKey changes above more efficient as they won't re-enumerate each time.

src/Directory/Identity/ApplicationUserPrincipal.cs
```diff
@@ -12,8 +13,9 @@ public class ApplicationUserPrincipal : ClaimsPrincipal
...
+        public IEnumerable<KeyValuePair<int, string>> Networks => FindAll(CustomClaimType.Network).Select(x => JsonConvert.DeserializeObject<KeyValuePair<int, string>>(x.Value));
```
beforan, line 17: Also probably should private helper method all of these. DeserializeJsonKeyValueClaims or something.

beforan (changes requested):

src/Directory/Biobanks.Web/Controllers/AccountController.cs
```diff
@@ -144,67 +143,78 @@ public async Task<ActionResult> Login(LoginModel model, string returnUrl)
...
             // if they have more than one claim, there's no obvious place to send them
-            if (CurrentUser.BiobankIds.Count() + CurrentUser.NetworkIds.Count() != 1)
+            if (CurrentUser.Biobanks.Count() + CurrentUser.Networks.Count() != 1)
                 return RedirectToAction("Index", "Home");
 
             //to returnUrl if appropriate, or a default route otherwise
             if (!string.IsNullOrEmpty(returnUrl) && Url.IsLocalUrl(returnUrl)) return Redirect(returnUrl);
 
-            // Single biobank only
+            // Biobank admin only
```
beforan, line 192: huh, this *is* single biobank only, because of line 186 which checks for multiple biobanks or networks xD

src/Directory/Biobanks.Web/Filters/AuthoriseToAdministerCapability.cs
```diff
@@ -17,16 +16,14 @@ public override void OnAuthorization(AuthorizationContext filterContext)
             base.OnAuthorization(filterContext);
 
             var session = filterContext.RequestContext.HttpContext.Session;
-            var activeOrganisationId = session[SessionKeys.ActiveOrganisationId];
+            var activeOrganisationId = Convert.ToInt32(session[SessionKeys.ActiveOrganisationId]);
 
-            if (activeOrganisationId != null 
-                && Convert.ToInt32(session[SessionKeys.ActiveOrganisationType]) == (int) ActiveOrganisationType.Biobank
-                && filterContext.RequestContext.HttpContext.User.ToApplicationUserPrincipal().BiobankIds.Contains(activeOrganisationId.ToString()))
+            if (Convert.ToInt32(session[SessionKeys.ActiveOrganisationType]) == (int)ActiveOrganisationType.Biobank
```
beforan, line 21: I hate this, but I don't think there's a better way, and I hate that there's not a better way.

Flemingjp (approved):

src/Directory/Biobanks.Web/Filters/AuthoriseToAdministerCapability.cs
beforan: You know what, there *is* a slightly better way, since this is functionality is used across multiple files and is ugly every time.

Write it into an enum extension method that compares the underlying value:

```csharp
public static class EnumExtensions {
	public static bool EnumEquals<TEnum> (this TEnum enumValue, string compareValue) where TEnum : struct  {
		if(!typeof(TEnum).IsEnum) throw new ArgumentException("Expected Enum type");
		try {
			var stringAsEnumValue = (TEnum)Enum.Parse(typeof(TEnum), compareValue);
			return enumValue.Equals(stringAsEnumValue);
		} catch (ArgumentException) {
			return false;
		}
	}
	public static bool EnumEquals<TEnum> (this TEnum enumValue, int compareValue) where TEnum : struct
		=> enumValue.EnumEquals(compareValue.ToString());
}
```

the above will do it for an int value, a string of the int value, or a string of the value name:

```csharp
MyEnum.Value.EnumEquals(1);
MyEnum.Value.EnumEquals("1");
MyEnum.Value.EnumEquals("Value");

will all work and return true/false appropriately
```

beforan: so in context, you can do:

```csharp
ActiveOrganisationType.Biobank.EnumEquals(
  session[SessionKeys.ActiveOrganisationType]);
```

beforan (changes requested):

src/Directory/Biobanks.Web/Utilities/SessionHelper.cs
```diff
@@ -14,7 +14,7 @@ public static int GetBiobankId(HttpSessionStateBase session)
             if (Convert.ToInt32(session[SessionKeys.ActiveOrganisationType]) == (int) ActiveOrganisationType.Biobank
```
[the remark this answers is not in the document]
beforan: can use new enum extension instead, which doesn't need to string convert

Flemingjp (approved):

beforan (approved):

PR #84, opened 2021-01-07
Add help links to registration pages

bradthompson995:
Links which can be configured have been added to the registration views which show up in a alert info box for the user to see.


src/Directory/Data/Constants/ConfigKey.cs
```diff
@@ -34,6 +34,8 @@ public static class ConfigKey
         public const string RegisterBiobankDescription = "site.register.biobank.description";
         public const string RegisterNetworkTitle = "site.register.network.title";
         public const string RegisterNetworkDescription = "site.register.network.description";
+        public const string EnableRegisterRegistrationHelpUrl = "site.register.enableregistrationhelpurl";
+        public const string RegisterRegistrationHelpUrl = "site.register.registrationhelpurl";
```
Flemingjp, line 39: Maybe have less verbose config keys ids? It not the only ones, but would stop adding anymore. My suggestions would be

`site.register.help.show` and
`site.register.help.url` 

Its a very minor point, but makes it more obvious the two keys are related

Flemingjp (changes requested):

Flemingjp (approved):

beforan (approved):

PR #88, opened 2021-01-18
Update Analytics to View only last 5 Quarters

shakisparki:
Options to select report period is turned off for individual organisations but available to ADAC users.
Tooltips on description of report options added for ADAC users, and bug fixes

### DIFF
--- a/src/Analytics/Services/GoogleAnalyticsReadService.cs
+++ b/src/Analytics/Services/GoogleAnalyticsReadService.cs
@@ -380,10 +380,9 @@ namespace Analytics.Services
             int month = endQuarter * monthsPerQuarter;
             int lastDayofQuarter = DateTime.DaysInMonth(year, month);
 
-            DateTimeOffset endDate = new DateTimeOffset(year, month, lastDayofQuarter,0,0,0,TimeSpan.Zero);
+            var endDate = new DateTimeOffset(year, month, lastDayofQuarter,0,0,0,TimeSpan.Zero);
             //get start date by subtracting report period (specified in quarters) from end date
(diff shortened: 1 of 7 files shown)


src/Analytics/Services/GoogleAnalyticsReadService.cs
```diff
@@ -383,6 +383,7 @@ public DateRange GetRelevantPeriod(int year, int endQuarter, int reportPeriod)
             DateTimeOffset endDate = new DateTimeOffset(year, month, lastDayofQuarter,0,0,0,TimeSpan.Zero);
```
Flemingjp, line 383: Use `var endDate` vs `DateTimeOffset`

src/Analytics/Services/GoogleAnalyticsReadService.cs
```diff
@@ -383,6 +383,7 @@ public DateRange GetRelevantPeriod(int year, int endQuarter, int reportPeriod)
...
             //get start date by subtracting report period (specified in quarters) from end date
             var startDate = endDate.AddMonths(-reportPeriod * monthsPerQuarter);
+            startDate = startDate.AddDays(1); //add +1 day as start date is included in report.
```
Flemingjp, line 387: Would it be clearer to use

```c#
var startDate = endDate.AddMonths(-1 * monthsPerQuarter * reportPeriod).AddDays(1);
```

The explicit -1 is more obvious / clearer and you might as well chain the transfoms

Flemingjp (changes requested):

Flemingjp (approved):

beforan (approved):

PR #89, opened 2021-01-20
Replace DeleteRefData method calls with API CRUD method

shakisparki:
A lot (if not all) refdata views call upon a custom Delete{RefData} function in the ADACController rather than their existing DELETE /api/{refdata} endpoint. This should be changed so the full CRUD functionality is used.

It may also be useful to have a general function (backend or frontend) for the sucess/failure feedback messages for ADAC RefData UI. As each RefData has its own fair of method in ADACController


src/Directory/Biobanks.Web/Controllers/ADACController.cs
```diff
@@ -778,6 +778,31 @@ public async Task<ActionResult> Analytics(int year = 0, int endQuarter = 0, int
 
         #region Reference Datasets
 
+        public ActionResult SetRefDataSuccessFeedbackAjax(RefDataFeedbackModel feedback)
```
shakisparki, line 781: Could also make this 3 functions for add, edit and delete and get rid of the switch case?

src/Directory/Biobanks.Web/Scripts/ADAC/adac-access-conditions.js
```diff
@@ -95,6 +94,18 @@ function AdacAccessConditionViewModel() {
...
+        // Send Feedback message
+        var url = "/ADAC/SetRefDataSuccessFeedbackAjax";
+        var data = {
+            name: name,
+            redirectUrl: redirectTo,
+            refDataType: refdata,
+            CRUDAction: action,
+        };
+        window.location.href = url + "?" + $.param(data);
+    }
```
Flemingjp, line 108: This could be moved to a new adac file as it will be used across all adac refdata scripts. Might as well try not adding any more duplication of code now

src/Directory/Biobanks.Web/Controllers/ADACController.cs
Flemingjp: Yepp feels like the clearer option. Could have three JavaScript functions to map across to as well.

src/Directory/Biobanks.Web/Views/ADAC/AccessConditions.cshtml
```diff
@@ -136,4 +134,5 @@
     @Scripts.Render("~/bundles/knockout")
     @Scripts.Render("~/bundles/datatables")
 	@Scripts.Render("~/bundles/adac/access-conditions")
+	@Scripts.Render("~/bundles/adac/refdata-feedback")
```
shakisparki, line 137: or include this in the respective refdata bundles e.g. ( ~/bundles/adac/access-conditions)

Flemingjp: Good question. Bundling per ADAC bundle would prevent any issues with missing reference - either way there's going to be some duplication of the definition. I would go with adding it to the individual bundles in the BundleConfig - will be easier than changing ever view.

src/Directory/Biobanks.Web/Scripts/ADAC/adac-access-conditions.js
```diff
@@ -106,10 +107,31 @@ $(function () {
...
-        bootbox.confirm("Are you sure you want to delete " + $link .data("access-condition") + "?",
+        bootbox.confirm("Are you sure you want to delete " + linkData.Description + "?",
             function (confirmation) {
-                confirmation && window.location.assign($link.attr("href"));
+                if (confirmation) {
+                    // Make AJAX Call
+                    $.ajax({
+                        url: url,
+                        type: 'DELETE',
+                        success: function (data, textStatus, xhr) {
```
beforan, line 120: can we factor this function out into a separate file, just like with `adac-refdata-feedback.js`? to avoid its repetition in EVERY ajax file. again.

beforan (changes requested):

beforan (approved):

Flemingjp (approved):

PR #83, opened 2021-01-07
Entities Project and Harmonization

Flemingjp:
Movement of Entities from the Directory and API into one shared Entities project.

 - [`93489bf`](https://github.com/biobankinguk/biobankinguk/pull/83/files/93489bf9751a0704e115d29247353c20edc2dabb) Solution / Project Restructuring
 - [`5a3a1a3`](https://github.com/biobankinguk/biobankinguk/pull/83/files/516a0251a6cf53ac315e269369f53b7dff529141..5a3a1a3d3eaa785987cd85a8955d78e7930539ac) Harmonization of Sex Entity
 - [`f075320`](https://github.com/biobankinguk/biobankinguk/pull/83/files/5a3a1a3d3eaa785987cd85a8955d78e7930539ac..0a308f64e746843d10187a0307c3db84c26d53df) Harmonization of PreservationType -> StorageTemperature
 - [`45d5c03`](https://github.com/biobankinguk/biobankinguk/pull/83/files/c0efb1a0af2116dd7e6c7a3562e4ccfc2151edd1..45d5c030e32b70071fe95070f0aa2bd752b2d849) Harmonization of MaterialType
 - [`28fd159`](https://github.com/biobankinguk/biobankinguk/pull/83/files/45d5c030e32b70071fe95070f0aa2bd752b2d849..28fd1596c1de8a1344f90fac3d661995659ec211) Harmonization of Diagnosis -> SnomedTerm
 - [`2aa77eb`](https://github.com/biobankinguk/biobankinguk/pull/83/files/c66902610c28334cb12056f7bcb6110ab9edbd3d..2aa77ebe988257c3861af7f3f21369870c1d3f51) Fixing View Model Referencing Build Errors

## Harmonization Changes Overview

Some of the changes in this PR include

- Sex
  - ADAC Sex RefData View
- Preservation Type
  - ADAC PreservationType RefData View
  - Add Collection Sample Set - `AddSampleSet.cshtml`
  - Edit Collection Sample Set - `EditSampleSet.cshtml`
  - Copy Collection Sample Set - `CopySampleSet.cshtml`
  - Elastic Search Facets
  - FacetList
- MaterialType
  -  MaterialTypeService
  - MaterialType RefData View
  - MaterialTypeGroup RefData View
- Diagnois/SnomedTerm
  - Id field now actual SNOMED ID, rather than some generated Diagnosis Id
  - Description [MaxLength:200] => No Limit
  - CapabilityDTO - Has Diagonis Field => SnomedTerm
[...]

PR #90, opened 2021-01-21
DataSeeder Upgrade

Flemingjp:
~~_**PR in draft as is dependent on #83. The relevant commits are [here](https://github.com/biobankinguk/biobankinguk/pull/90/files/41b763a2eb948da11c39c919e4b772e43e120a96..cef8228b01a9f0a6c4c4e265f02f420907f61c31)**_~~

Upgraded the DataSeeder to update csv's of harmonized entities, and to include seeding for the newly introduced API entities.

Other upgrades include

- The `DbSet` for a given entity is now inferred directly from the generic `db.Set<TEntity>()` vs `db.Set(Type type)` removing the need for a lookup table of type and table name
- The loaded csv data is also uses `csv.GetRecords<T>()` vs `csv.GetRecords(Type type)` for the same reasons
- Unified all seed actions into one list, with the seeding order determined by the list order. This removes the need for the separate `LeaveTilLast` list of seed actions. 
- `SeedingService` now uses the logger vs `Console.WriteLine` and all logging is done via the `Seed<T>` method such that all seeding is properly logged
- Updated the sample-seed-data to reflect the changes to the entities fields and names


src/Entities/Api/ReferenceData/OntologyVersion.cs
```diff
@@ -27,7 +27,6 @@ public class OntologyVersion
         /// <summary>
         /// Ontology to which the version relates.
         /// </summary>
-        [Required]
-        public Ontology Ontology { get; set; }
```
Flemingjp, line 31: Removal of `[Required]` as would throw an error when trying to insert any entity into the table with a `null` value - not sure of the implications to the API

beforan: it should be required. How can you have an OntologyVersion that doesn't relate to an Ontology?

better than breaking the model to suit your data would be fixing your data to match the model ;)

In this case, why not insert the only Ontology, and version, we support today, to fulfill the API's expectations.

As per the current QA API database:

Ontology: SNOMED-CT
Ontology Version: Value - 25

src/Directory/DataSeed/Services/SeedingService.cs
```diff
@@ -22,209 +23,193 @@ namespace Biobanks.DataSeed.Services
...
+            return _db.Configs.FirstOrDefault(x => x.Key == key).Value;
+        }
+
+        private IEnumerable<T> ReadCsv<T>(string filePath = "") where T : class
+        {
+            if (string.IsNullOrEmpty(filePath))
             {
-                IgnoreReferences = true
-            };
+                filePath = $"{ _dataDir }/{ typeof(T).Name }.csv";
```
beforan, line 205: use `Path.Combine()`

src/Entities/Api/ReferenceData/OntologyVersion.cs
beforan: I see we have data for those, so I'm not sure what the issue is? when are we trying to insert null FKs?

beforan (changes requested):

beforan (approved):

JosephLavagna (approved):

PR #94, opened 2021-01-26
Annotation batch job

JosephLavagna:
- Pulls all publications from the directory 
- Gets annotations for these publications from the Epmc API
- Pushes these annotations to a new annotations table in the db

-Many to many relationship between Annotations and Publications
- Manually created join table in Publications solution since using EF Core 3.1
- Replicated files/methods for Publications batch job for annotations
- Annotation batch job runs alongside publications in BatchFunction


src/Directory/Data/Migrations/202101261125598_AddAnnotationIdColumnToAnnotations.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    using System;
+    using System.Data.Entity.Migrations;
+    
+    public partial class AddAnnotationIdColumnToAnnotations : DbMigration
+    {
+        public override void Up()
+        {
+            AddColumn("dbo.Annotations", "AnnotationId", c => c.String());
+        }
```
Flemingjp, line 11: This can be merged with the initial Annotations migration

src/Publications/Publications/Entities/PublicationAnnotation.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Publications.Entities
+{
+    public class PublicationAnnotation
+    {
+        public int Publication_Id { get; set; }
```
Flemingjp, line 9: Better to have `Publication_Id` => `PublicationId` and `Annotation_Id` => `AnnotationId` ?

Flemingjp (changes requested):

src/Publications/Publications/Entities/PublicationAnnotation.cs
JosephLavagna: Basically since EF6 (Directory) automatically creates the join entity - It created the table with those column names - and since I had to manually create this join entity in the Publications solution - EF Core 3.1 it had to match those column names or it would throw an error when pushing to the database. I am not sure if theres a way around this?

JosephLavagna: And those Publication_Id/Annotation_Id are the Id values given by EF, but PublicationId and AnnotationId are both values from the API so may be best to leave it how it is to avoid confusion?

Flemingjp (approved):

beforan (approved):

PR #99, opened 2021-02-01
Allow publications to be disabled on a biobank level

shakisparki:
Some Biobanks, such as AZ, have a huge number of publications due to their biobank being labelled with the company name. However, most of these publications will be matched based on name, so AZ won't want to have to go through this. Therefore, we need an option wtihin a biobank admin section to turn off publications.


src/Directory/Entity/Data/Organisation.cs
```diff
@@ -54,6 +54,8 @@ public class Organisation
 
         public string OtherRegistrationReason { get; set; }
 
+        public bool? IncludePublications { get; set; }
```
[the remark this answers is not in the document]
beforan: Actually I guess `null` is equivalent to `true` since that's what all your null propagation code defaults to.

Either way, in the domain, this is a true boolean, not a nullable boolean. There are two possible values, `true` or `false`. `null` is not applicable here.

Because including is the default state, however, we need to either set default to `true` in the db schema, or switch the field so it is `ExcludePublications` (so that `false` can be default).

This is cleaner, safer, better represents the intent, safes lots of null propagation code which if forgotten could cause bugs... etc...

src/Publications/Publications/Services/Hosted/FetchPublicationsService.cs
```diff
@@ -39,7 +39,7 @@ public class FetchPublicationsService : IHostedService
         public async Task StartAsync(CancellationToken cancellationToken)
         {
             // Call directory for all active organisation
-            var biobanks = await _biobankReadService.ListBiobanksAsync();
+            var biobanks = (await _biobankReadService.ListBiobanksAsync()).Where(x => x.IncludePublications != false).ToList();
```
beforan, line 42: if the property becomes a regular boolean, this comparison is simpler: `x => x.IncludePublications`

src/Directory/Biobanks.Web/Views/Biobank/Publications.cshtml
```diff
@@ -11,15 +11,21 @@
...
-			Approving publications will help people find you by your research.
-		</div>
+        <div class="alert alert-info feedback-message annualstats-feedback" role="alert">
+            Approving publications will help people find you by your research.
+        </div>
 
-		<table id="biobank-publications" class="table table-striped table-hover" style="width:100%"></table>
-	</div>
+        @*Enable/Disable Publications*@
+        <div class="btn-group btn-toggle pull-right">
```
beforan, line 22: Is there a good reason we can't use a checkbox for this?

Elsewhere in the site we do for boolean toggles.

Also saves us doing all the visual state management ourselves in js. less code  = better.

src/Directory/Biobanks.Web/Scripts/Biobank/publications.js
```diff
@@ -142,4 +142,54 @@ $(function () {
...
+                togglePubBtn("off");
+                $('#biobank-publications_wrapper').hide();
+            }
+        }
+    });
+});
+
+//toggle enable/disable publications button
+function togglePubBtn(val) {
+    var $onBtn = $("#on-btn");
```
beforan, line 161: 90% of this code goes away if you use a checkbox.

src/Directory/Biobanks.Web/Scripts/Biobank/publications.js
```diff
@@ -142,4 +142,54 @@ $(function () {
...
+        $btnGroup.find('.btn').toggleClass('active');
+        $btnGroup.find('.btn').toggleClass('btn-primary');
+        $btnGroup.find('.btn').toggleClass('btn-default');
+        return $($btnGroup.find('.active')).data("incl-publications");
+    }
+}
+
+$('.btn-toggle').click(function () {
+    var btnVal = togglePubBtn();
+    if (btnVal)
```
beforan, line 190: I don't particularly like the show/hide here, but i guess it makes the most sense on the client side in a pre virtual dom world.

src/Directory/Biobanks.Web/Controllers/BiobankController.cs
```diff
@@ -1777,6 +1777,43 @@ public async Task<JsonResult> ClaimPublicationAjax(string publicationId, bool ac
                 });
             }
         }
+
+        [HttpGet]
+        [Authorize(ClaimType = CustomClaimType.Biobank)]
+        public async Task<JsonResult> IncludePublicationsAjax()
```
beforan, line 1783: can we not add any new `-Ajax` methods to web controllers that put a `success` boolean in the response body? Ideally over time we would work to eliminate the ones we already have, not add more!

The whole point of moving all the refdata admin stuff to api controllers was to avoid this. so let's continue onwards using api controller endpoints for ajax interactions.

I guess these methods will warrant a new Api Controller, for general purpose Biobank related ajax calls

They're not really RESTful, more like an RPC for a biobank setting, but that's fine. I think the route would be something like:

`/api/biobank/includepublications`
- with a GET for the current value (`true` or `false`)
- PUT to set the value with a route parameter or body payload of `true` or `false`.

Obsiously the AJAX call changes so it error handles properly on HTTP status, not the success field. also the not very clearly named "status" property is eliminated as it's a single value endpoint.

beforan (changes requested):

src/Directory/Biobanks.Web/Views/Biobank/Publications.cshtml
shakisparki: Ok, could you point me to where a checkbox was used?

beforan: https://github.com/biobankinguk/biobankinguk/search?l=HTML%2BRazor&q=type%3D%22checkbox%22

beforan (approved):

Flemingjp (approved):

JosephLavagna (approved):

PR #100, opened 2021-02-05
Annotations Batch Job - Type Filter on API Call

JosephLavagna:
- Filters Annotations from API by list of types
- Fixed duplicate annotation check before pushed to database


src/Publications/Publications/Services/EpmcWebService.cs
```diff
@@ -93,7 +94,16 @@ private async Task<List<AnnotationDto>> AnnotationSearch(string publicationId, s
...
-            string endpoint = QueryHelpers.AddQueryString("annotations_api/annotationsByArticleIds", parameters);
+            // Filter by type of Annotation
+            var types = new List<string>();
+            types.Add("Diseases");
+            types.Add("Organ Tissue");
+            types.Add("Phenotype");
+            types.Add("Sample-Material");
+            types.Add("Body-Site");
+            string typeQuery = "?type=" + string.Join("&type=", types);
+
```
Flemingjp, line 105: Can initialize the list directly with the values. 

```c#
var types = new List<string>()
{
    "Diseases",
    "Organ Tissue"
};
```
Should be able to add the `type` query to the parameters dictionary for the QueryHelper to parse automatically. I'm assuming current solution is a workaround for EMPC use of multiple `type` and QueryHelpers lack of support for it?

```c#
var parameters = new Dictionary<string, string>()
{
    { "articleIds", $"{source}:{publicationId}" },
    { "format", "JSON" },
    { "type", string.Join("&type=", types) }
};
```

Flemingjp (changes requested):

src/Publications/Publications/Services/EpmcWebService.cs
beforan: or use [Flurl](https://flurl.dev/docs/fluent-url/)

notably:
> Url.QueryParams is a special collection type that maintains order and allows duplicate names, but is optimized for the typical case of unique names

JosephLavagna: yep exactly - epmc uses multiple 'type' for its endpoint and QueryHelper only supports the use of a dictionary. Unfortunately that solution doesn't return any annotations.

src/Publications/Publications/Services/AnnotationService.cs
```diff
@@ -28,30 +28,32 @@ public async Task AddPublicationAnnotations(string publicationId, IEnumerable<An
...
+                          Publication_Id = existingPublications.Id
+                      };
+                      annotationEntity.PublicationAnnotations.Add(publicationAnnotation);
+                      annotationList.Add(annotationEntity);
+
                 }
             }
-
+            //Remove duplicate Annotation Names
+            var annList = annotationList.GroupBy(x => x.Name).Select(x => x.First()).ToList();
```
beforan, line 53: `ToHashSet().ToList()` will do this but look less complex ;) (and the framework is probably more efficient than you ;))

Also, does it need to be a list? could just `ToHashSet()` and then iterate over that instead of a list?

Yes, HashSet is my new favourite type. Yes, it's your fault.

src/Publications/Publications/Services/AnnotationService.cs
```diff
@@ -28,30 +28,32 @@ public async Task AddPublicationAnnotations(string publicationId, IEnumerable<An
...
+   
             //Add or Update new annotations
-            foreach (var newer in annotationList)
+            foreach (var newer in annList)
             {
                 //Find if older version of annotation exists
                 var older = existingAnnotations.Select(x => x.Annotation).Where(a => a.AnnotationId == newer.AnnotationId).FirstOrDefault();
```
beforan, line 59: you can inline this. Most LINQ selectors like `First()` accept filter expressions.

sooo: `existingAnnotations.Select(x => x.Annotation).FirstOrDefault(a => a.AnnotationId == newer.AnnotationId);`

beforan (changes requested):

src/Publications/Publications/Services/AnnotationService.cs
beforan: actually, your grouping on a property so it might be more complicated than i thought...

Flemingjp (approved):

beforan (approved):

PR #106, opened 2021-02-09
Add PreservationType to Samples The API

Flemingjp:
Implementing PreservationType on the API side including validation. Some minor bug fixes that were required to get the API working after moving to the central Biobanks database


src/Api/SubmissionAzureFunction/Services/ReferenceDataReadService.cs
```diff
@@ -155,6 +155,13 @@ public async Task<StorageTemperature> GetStorageTemperature(string value)
             => (await ListStorageTemperatures())
                 .FirstOrDefault(x => x.Value.Equals(value, StringComparison.CurrentCultureIgnoreCase));
 
+        public async Task<IEnumerable<PreservationType>> ListPreservationTypes() 
+            => await _db.PreservationTypes.AsNoTracking().ToListAsync();
```
Flemingjp, line 159: Is caching required - how does this work out if the reference data is changed on the Directory side?

beforan: > how does this work if the [...] data is changed

I mean, that is the million dollar question in all caching scenarios.

I don't think we need caching of this data yet.

If and when we do, we will look at appropriate strategies for invalidation on source data change, or "eventual consistency", but that's for another day.

Flemingjp: PBI 30996

src/Data/BiobanksDbContext.cs
```diff
@@ -70,7 +71,7 @@ public class BiobanksDbContext : DbContext
         public DbSet<Publication> Publications { get; set; }
 
         /* API */
-        public DbSet<Status> Statuses { get; set; }
+        public DbSet<Status> Status { get; set; }
```
Flemingjp, line 74: Biobanks table is currently called `Status` vs `Statuses` therefore was causing errors. Might be nicer to rename the table to `Statuses` as all other tables are plural.

src/Api/SubmissionAzureFunction/Services/ReferenceDataReadService.cs
```diff
@@ -123,7 +123,9 @@ public async Task<IEnumerable<MaterialType>> ListMaterialTypes()
 
         public async Task<MaterialType> GetMaterialTypeWithGroups(string value)
             => (await ListMaterialTypes())
-                .FirstOrDefault(x => x.Value.Equals(value, StringComparison.CurrentCultureIgnoreCase));
+                .FirstOrDefault(x => 
+                    (x.Value.Equals(value, StringComparison.CurrentCultureIgnoreCase)) &&
+                    (x.MaterialTypeGroups?.Any() ?? false));
```
Flemingjp, line 128: Added condition to ensure there are some MaterialTypeGroups. It wasn't doing this check, causing NPE in several places.

src/Api/SubmissionAzureFunction/Services/SampleWriteService.cs
```diff
@@ -53,7 +53,7 @@ public async Task ProcessSamples(IEnumerable<SampleDto> dto)
                 .Where(x => x.OrganisationId == organisationId)
                 .AsNoTracking()
                 .ToListAsync();
-            var allLiveSamples = await _db.Samples
+            var allLiveSamples = await _db.LiveSamples
                 .Where(x => x.OrganisationId == organisationId)
```
Flemingjp, line 57: Same issue with `Statuses`

src/Data/BiobanksDbContext.cs
beforan: yes please to the table rename.

src/Api/SubmissionAzureFunction/Services/SampleWriteService.cs
```diff
@@ -122,6 +122,10 @@ public async Task ProcessSamples(IEnumerable<SampleDto> dto)
 
                     continue;
                 }
+                catch
```
beforan, line 125: This is almost certainly not the right thing to do

Why are we catching all exceptions in the above block and then doing nothing?

- Can we catch only specific actions?
- What should we do instead of nothing?

src/Api/SubmissionAzureFunction/Services/SampleWriteService.cs
beforan: same issue?

beforan (changes requested):

src/Api/SubmissionAzureFunction/Services/SampleWriteService.cs
Flemingjp: Not sure what has changed here? I'll check and get back to you

src/Api/SubmissionAzureFunction/Services/SampleWriteService.cs
Flemingjp: Well similar, the table is called `LiveSamples` where the context was expecting `Samples`

beforan (approved):

JosephLavagna (approved):

PR #107, opened 2021-02-09
Annotations batch job fix

JosephLavagna:
- Function was crashing when it was given a publication with no source string.
- Added a null check for publication id and source.
- Added a try/catch for the API call itself.


src/Publications/Publications/Services/EpmcWebService.cs
```diff
@@ -109,16 +118,24 @@ private async Task<List<AnnotationDto>> AnnotationSearch(string publicationId, s
 
             var url = new Url("annotations_api/annotationsByArticleIds");
             url.SetQueryParams(parameters).SetQueryParam("type", types);
+            var result = new List<AnnotationResult>();
 
+            try 
+            {
+                string response = await _client.GetStringAsync(url);
```
blargly, line 125: var

blargly (changes requested):

Flemingjp (approved):

blargly (approved):

PR #110, opened 2021-02-11
MaterialType Dataseeder Fixes

Flemingjp:
- Fixes issue with MaterialTypes/MaterialTypeGroup join table
- Seeds MaterialTypes properly to ensure no duplication of MaterialTypeGroups


src/Directory/DataSeed/Services/SeedingService.cs
```diff
@@ -110,6 +110,23 @@ private void SeedCountries()
...
+
+            Seed(
+                ReadJson<MaterialType>()
+                    .Select(x => new MaterialType()
+                    {
+                        Value = x.Value,
+                        SortOrder = x.SortOrder,
+                        MaterialTypeGroups = x.MaterialTypeGroups
+                            ?.Select(y => validGroups.First(z => z.Value == y.Value))
+                            .ToList()
```
Flemingjp, line 125: Alternative logic can pull from the known groups, rather than matching to known groups. This way will not throw an error, instead will just drop unrecognised MaterialTypeGroups. Not sure what the preferred behaviour should be.

```c#
MaterialTypeGroups = validGroups
    .Where(y => x.MaterialTypeGroups?.Any(z => z.Value == y.Value) ?? false)
    .ToList()
```

blargly: I'm thinking an obvious error might be best during seeding, or if we were to drop MaterialTypeGroups, we need to make sure its logged for the end user to clearly seel

Flemingjp: Agreed 👍

blargly (approved):

JosephLavagna (approved):

PR #116, opened 2021-02-16
A: Improve search typeahead experience to accomodate "other terms"

shakisparki:
currently, typeahead suggests (autocompletes) the primary name of a disease status.
Once other terms are present, we should only show one suggestion per primary name (as now), but have a richer results view, something like:

----
Primary name
(other term being autocompleted)
(...other possible terms)
----

This PR doesn't seperate out the other terms being completed from the rest of the other terms (into a separate row). However the other term is highlighted as the user types. So we have:

----
Primary name
(...other terms being autocompleted)
----

image

Compare with #117

### DIFF
--- a/src/Directory/Biobanks.Web/Content/Site/Site.css
+++ b/src/Directory/Biobanks.Web/Content/Site/Site.css
@@ -1079,6 +1079,11 @@ div.info div {
     cursor: pointer;
 }
 
+.search-list{
+    border-bottom: 0.5px solid lightgrey; 
+    text-align: left;
+    font-size:medium;
+}
 
 /*Toggle switch*/
 /* The switch - the box around the slider */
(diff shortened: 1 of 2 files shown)


Flemingjp:
If we are voting between A/B - I prefer this one, more compact and the the bold text makes it obvious where the matching is. A further suggestion would be to remove the whitespace above the primary name.

JosephLavagna:
> If we were to vote between A/B - I prefer this one, more compact and the the bold text makes it obvious where the matching is. A further suggestion would be to remove the whitespace above the primary name.

I second this

Flemingjp:
@shakisparki @beforan @blargly Close this PR if we're not going with this style?

beforan:
closing in favour of #117

PR #117, opened 2021-02-16
B: Improve search typeahead experience to accomodate "other terms"

shakisparki:
currently, typeahead suggests (autocompletes) the primary name of a disease status.

Once other terms are present, we should only show one suggestion per primary name (as now), but have a richer results view, something like:

----
Primary name
(other term being autocompleted)
(...other possible terms)
----

This PR presents the results in the exact way as shown above at the cost of extra logic. See img below

image

Compare with #116


src/Directory/Biobanks.Web/Controllers/SearchController.cs
```diff
@@ -209,15 +209,17 @@ public async Task<JsonpResult> SearchOntologyTerms(string searchDocumentType, st
 
         private async Task<List<OntologyTermModel>> GetOntologyTermSearchResultsAsync(SearchDocumentType type, string wildcard)
         {
-            var ontologyTerms = await _biobankReadService.ListSearchableOntologyTermsAsync(type, wildcard);
+            var ontologyTerms = _mapper.Map<IEnumerable<OntologyTermResult>>(await _biobankReadService.ListSearchableOntologyTermsAsync(type, wildcard));
```
Flemingjp, line 213: Since this result is then transformed to a OntologyTermModel straight afterwards - is the AutoMapper a bit redundant. Should be able to map directly to an OntologyTermModel manually?

src/Directory/Services/BiobankReadService.cs
```diff
@@ -1205,11 +1205,20 @@ public async Task<bool> GetSiteConfigStatus(string siteConfigValue)
...
+                listOntologyTerms, o => o.Value.ToLower(), i => i.OntologyTerm, (a, b) => new OntologyTermResultDTO
+                {
+                    OtherTerms = a.OtherTerms ?? "",
+                    MatchingOtherTerms = b.MatchingOtherTerms,
+                    Id = a.Id,
+                    Value = a.Value,
+                    NonMatchingOtherTerms = a.OtherTerms?.Split(',').Select(p => 
+                        p.Trim()).Where(m => !(b.MatchingOtherTerms.Contains(m))).ToList() ?? new List<string>()
+                });
         }
```
Flemingjp, line 1222: Any chance of splitting this code up for the sake of clarity. There's a lot going on in one place, and it should be as easy to read where possible.

src/Directory/Search/Elastic/BaseElasticSearchProvider.cs
```diff
@@ -213,5 +214,31 @@ void AddAgg(string name)
...
+                        var hl = terms.Highlight["ontologyOtherTerms.name"].First();
+                        if (!ontologyTerm.MatchingOtherTerms.Any(x => x == hl))
+                            ontologyTerm.MatchingOtherTerms.Add(hl);
+                    }
+                }
+                ontologyTerms.Add(ontologyTerm);
+            }
+
+            return ontologyTerms;
+        }
```
Flemingjp, line 242: Same with clarity here - probably can make it clearer by trying to use LINQ to reduce the three explicit for loops. Thinking something like this (untested idea).

```c#
protected static IEnumerable<OntologyTermsSummary> ExtractOntologyOtherTermsHits(ISearchResponse<BaseDocument> searchResponse)
{
    var ontologyTerms = searchResponse.Aggregations
            .Terms("diagnoses")
            .Buckets
            .Select(x => 
                new OntologyTerm
                {
                    OntologyTerm = x.Key,
                    MatchingOtherTerms = new ();
                }
            );

    foreach (var ontologyTerm in ontologyTerms)
    {
        searchResponse.Hits
            .Where(x => x.Source.OntologyTerm.ToLower() == ontologyTerm.OntologyTerm)
            .Where(x => 
                !ontologyTerm.MatchingOtherTerms.Any(y => 
                    y == terms.Highlight["ontologyOtherTerms.name"].FirstOrDefault()
                )
            )
            .ToList()
            .ForEach(ontologyTerm.MatchingOtherTerms.Add)
    }

    return ontologyTerms;
}
```

I think clarity in the search project should be priority as a lot of the code is already unknown and unclear.

Flemingjp (changes requested):

src/Directory/Search/Elastic/BaseElasticSearchProvider.cs
shakisparki: In terms of clarity, I think having explicit foreach statement does that better than compact LinQ statements especially for the complex search queries. In terms of compactness of code, I don't seem to be able to get your sample to work correctly. If you can get a working code then please do forward and I will happily replace the block.

beforan (approved):

blargly (approved):

Flemingjp (approved):

PR #118, opened 2021-02-17
Add ApplicationInsights Logging On SuperUser Search Controller

Flemingjp:
- Added logging to SuperUser SearchIndex controller endpoint
- Added additional feedback to user on frontend


blargly (approved):

src/Directory/Services/BiobankIndexService.cs
```diff
@@ -52,8 +52,7 @@ public async Task<string> GetClusterHealth()
             }
             catch
             {
-                // Exception Occurred - Assume Search Is Down
-                return "red";
+                return null;
```
beforan, line 55: we don't track this exception? should we?

just feels inconsistent, and it's very presumptuous to say any exception is search is down, which makes me think we should track it so we have visibility of exceptions that aren't just "search is down" if and when they occur.

beforan (changes requested):

src/Directory/Services/BiobankIndexService.cs
Flemingjp: Can be done - in that case is it worth to use the same AppTelemetryClient - rather than creating a new one per error thrown?

beforan (approved):

blargly (approved):

PR #119, opened 2021-02-17
All RefData API Endpoint Work Without Authentication

shakisparki:
Any of the Reference Data endpoints can be used without having the ADAC Claim, therefore any user can add, change and delete reference data values. All API Endpoint except for the read endpoints (GET) should require the ADAC UserClaim to ensure only ADAC Users can change the data.

Accessing all write endpoints on the Reference Data APIControllers without the correct ADAC UserClaim should return HTTP 403 Forbidden response (or equivalent)

All users, regardless of UserClaim, should be able to read all data from the APIControllers (mainly all GET endpoints)

### DIFF
--- a/src/Directory/Biobanks.Web/Filters/UserApiAuthorizeAttribute.cs
+++ b/src/Directory/Biobanks.Web/Filters/UserApiAuthorizeAttribute.cs
@@ -1,15 +1,31 @@
 ﻿using System.Web.Http.Controllers;
 using System.Net.Http;
 using System.Net;
+using System.Web.Http;
+using System.Linq;
 
 namespace Biobanks.Web.Filters
 {
     public class UserApiAuthorizeAttribute: System.Web.Http.AuthorizeAttribute
     {
-        //override the HTTP Unauthorized handler to return a 403 Forbidden, instead of 401 Unauthorized
-        protected override void HandleUnauthorizedRequest(HttpActionContext filterContext)
+        public override void OnAuthorization(HttpActionContext filterContext)
         {
-            filterContext.Response = new HttpResponseMessage(HttpStatusCode.Forbidden);
+            //Skip Authentication for AllowAnonymous decorated contexts (Action or Controller, with inheritance)
+            if(filterContext.ActionDescriptor.GetCustomAttributes<AllowAnonymousAttribute>(true).Any() ||
+                filterContext.ActionDescriptor.ControllerDescriptor.GetCustomAttributes<AllowAnonymousAttribute>(true).Any())
+                return;
+
+            //Check that the user is authenticated
+            if (!filterContext.RequestContext.Principal.Identity.IsAuthenticated)
+            {
+                filterContext.Response = new HttpResponseMessage(HttpStatusCode.Unauthorized); //401 Unauthorised
+                return;
+            }
+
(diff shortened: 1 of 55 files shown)


Flemingjp (changes requested):
Has the Filter been added to the FilterConfig? It won't be in use if not?

shakisparki:
> Has the Filter been added to the FilterConfig? It won't be in use if not?

It will be if decorated per controller. Adding it to filterConfig will apply it to all controllers.

Flemingjp (approved):

src/Directory/Biobanks.Web/Filters/UserApiAuthorizeAttribute.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Net;
+
+namespace Biobanks.Web.Filters
+{
+    public class UserApiAuthorizeAttribute: System.Web.Http.AuthorizeAttribute
+    {
+        //override the HTTP Unauthorized handler to return a 403 Forbidden, instead of 401 Unauthorized
+        protected override void HandleUnauthorizedRequest(HttpActionContext filterContext)
+        {
+            filterContext.Response = new HttpResponseMessage(HttpStatusCode.Forbidden);
```
beforan, line 12: This is incorrect behaviour.

I think you've copied it from the equivalent MVC filter, but that one works because it assumes we have the global filters set up the way we do. (Pretty opaque, I know, and not at all documented - I had to root around to confirm it).

So here, we are erroneously returning 403 for all failure cases. Really we should check if the user is authenticated and return 401 if they aren't or 403 if they are authenticated but aren't authorised. I think we can do it all in the one filter, assuming the context can tell us if the user is authenticated in this handler?

Optional detailed essay on MVC5's broken auth filters:

So in MVC:

- We have a global filter that checks authentication on all requests that aren't exempted through `AllowAnonymous`.
- That filter never checks authorization (roles etc.)
- It uses `401 Unauthorized` on failure, which is correct (we had to correct this from default MVC5 behaviour).
- Then we have a non-global Authorization filter, derived from the default MVC AuthorizeAttribute.
- The MVC5 AuthorizeAttribute checks the user is Authenticated AND that they have specified Roles.
- It then erroneously returns 401 Unauthorised for any authorisation failure.
- It *should* return 403 Forbidden for cases where the user *is* authenticated but isn't in the correct role.
- Sooo our custom Authorization Filter returns 403 Forbidden in all cases. It can do this, because it knows if we got this far, we passed the Authentication Filter successfully. Therefore it knows we are authenticated, and won't be needing a 401. So it can return 403 for all remaining cases.

So it's the combination of the authentication and authorization filters in MVC that makes it work correctly.
It's confusing because one is global, and the other isn't, and they both do the authentication check :\ But that's MVC5's fault (and is resolved in Core).

beforan (changes requested):

src/Directory/Biobanks.Web/Filters/UserApiAuthorizeAttribute.cs
shakisparki: Thank you very much. This makes it very clear.

Flemingjp (approved):

beforan (approved):

PR #121, opened 2021-02-18
Publications on/off toggle is missing a label

shakisparki:
In an individual biobank, the toggle for switching on publications has no label to make it clear what the purpose of the toggle is or what the current publication status is.

PR: Adds label and hover tooltip


Flemingjp (commented):
Just to check - This further worsen the issue with the slider css?

beforan:
> Just to check - This further worsen the issue with the slider css?

I wouldn't expect it to worsen it? the added CSS in this PR uses different classes not already used elsewhere in the site's markup.
In which case that issue stands as a separate bug and can have a separate PR.

The class names are still very general though. I really would rather prefer (and it would be more accessible) if we just used a plain checkbox with a label (which would then function as a hit area for the toggle as well).

Flemingjp (dismissed):

beforan (approved):

Flemingjp (approved):

PR #122, opened 2021-02-18
Superuser - Build Index Functionality

bradthompson995:
The configure-search script from the elastic-search directory has been integrated into the superuser directory in the form of a Build Index button which deletes the index and the builds the index from the supplied JSON files, which are stored in the 'App_Config' folder within the web directory with the other JSON files.  

It also then calls the existing rebuild index function to populate the index.


src/Directory/Services/BiobankIndexService.cs
```diff
@@ -34,6 +36,61 @@ public class BiobankIndexService : IBiobankIndexService
...
 
+        public async Task BuildIndex()
+        {
+            //Building the Search Index
+
+            var searchBase = ConfigurationManager.AppSettings["ElasticSearchUrl"];
+            List<string> _navPaths = new List<string>();
+            _navPaths.Add(HostingEnvironment.MapPath(@"~/App_Config/capabilities.json"));
+            _navPaths.Add(HostingEnvironment.MapPath(@"~/App_Config/collections.json"));
+
```
Flemingjp, line 47: Can be simply written as

```c#
var _navPaths = new List<string>()
{
    HostingEnvironment.MapPath(@"~/App_Config/capabilities.json"),
    HostingEnvironment.MapPath(@"~/App_Config/collections.json")
};
```

src/Directory/Services/BiobankIndexService.cs
```diff
@@ -34,6 +36,61 @@ public class BiobankIndexService : IBiobankIndexService
...
+                    }
+                
+                    using (var client = new HttpClient())
+                    {
+                        client.BaseAddress = new Uri(searchBase);
+
+                        //Creating the Index                      
+                        HttpContent pathContent = new StringContent(System.IO.File.ReadAllText(path), System.Text.Encoding.UTF8, "application/json");
+                        var response = await client.PutAsync($"{searchBase}/{fileName}", pathContent);
+                    }
```
Flemingjp, line 67: Not sure if you really need to use two seperate HttpClients. You shouldnt need to set the base address either if you're using the the searchBase in the Delete/Put calls.

src/Directory/Services/BiobankIndexService.cs
```diff
@@ -34,6 +36,61 @@ public class BiobankIndexService : IBiobankIndexService
...
+                }
+                catch
+                {                  
+                    return;
+                }
```
Flemingjp, line 72: @beforan will probably want some logging to Application Insights here 😉. Saying that @beforan if we were to start more extension AI logging, would it be worth writing a AILogger and using that throughout?

```c#
var ai = new TelemetryClient();
ai.LogError() // Something to that effect
```

Flemingjp (commented):
Looking good so far - comments are just suggestions you may as well have whilst working on this PR actively

Flemingjp (dismissed):

src/Directory/Biobanks.Web/Controllers/SuperUserController.cs
```diff
@@ -86,6 +86,16 @@ public async Task<RedirectToRouteResult> ReindexAllData()
 
             return RedirectToAction("SearchIndex");
         }
+
+        public async Task<RedirectToRouteResult> BuildIndex()
+        {
+            await _indexService.BuildIndex();
```
beforan, line 92: If any handled exceptions (http or file) occurred, we can set a failure feedback message appropriately, and not proceed with the rest of this method

src/Directory/Biobanks.Web/App_Config/capabilities.json
```diff
@@ -0,0 +1,73 @@
+﻿{
```
beforan, line 1: these files are duplicated from the ones in the elastic search folder, now?

Having them in 2 places in the repo will be a nightmare to keep in sync if they change, and they don't really belong to the directory, they are just used by it.

I think they should be put in place possibly when publishing. Can the csproj reference the ones on disk higher up? (I don't see why not). If not, then we can copy them in as content in the deployment pipeline.

beforan (changes requested):

Flemingjp (approved):

beforan (approved):

PR #124, opened 2021-02-19
Move Migrations To EFCore DbContext

Flemingjp:


src/DataSeed/Transforms/UrlTransformer.cs
```diff
@@ -0,0 +1,45 @@
...
+            if (!Uri.TryCreate(urlString, UriKind.Absolute, out var url))
+            {
+                return false;
+            }
+
+            // Check For Mistyped Scheme -> Manual Parsing
+            if (validSchemes.Any(s => url.Host == s))
+            {
+                url = new Uri($"{url.Host}:{url.PathAndQuery}", UriKind.Absolute);
+            }
```
Flemingjp, line 34: @beforan Am I right in thinking that since the scheme is mistyped, the actual scheme has been parsed as the Host and the Host as the PathAndQuery?

src/DataSeed/Data/DirectoryConfigs.cs
```diff
@@ -0,0 +1,280 @@
...
+{
+    public static class DirectoryConfigs
+    {
+        public static readonly List<Config> DefaultConfigs = new List<Config>
+        {
+            #region Display
+            // Show About Page
+            new ()
+            {
+                Key = "site.display.aboutpage",
```
beforan, line 14: I think all this content belongs specifically to the Directory app, so it should probably be responsible for ensuring it's present. Probably on app startup.

src/DataSeed/Transforms/UrlTransformer.cs
beforan: I can't remember the exact issue, but I don't think this data fix is really suitable being in the seeder.

It's more like a utility for if we encoutner broken URLs. Good times to run it might be on Directory startup, or we could trigger it from the SuperUser panel instead. Just as another piece in that toolkit.

beforan (changes requested):

src/Directory/Data/BiobanksDbInitializer.cs
```diff
@@ -0,0 +1,74 @@
...
+    public class BiobanksDbInitializer : IDatabaseInitializer<BiobanksDbContext>
+    {
+        public void InitializeDatabase(BiobanksDbContext context)
+        {
+            SeedDirectoryConfigs(context);
+            SeedOrganisationTypes(context);
+            FixOrganisationUrls(context);
+        }
+
+        private void FixOrganisationUrls(BiobanksDbContext context)
```
Flemingjp, line 20: Do we want this as SuperUser tool instead? @beforan @blargly

beforan: not a bad shout at all. I'm up for it

blargly (approved):

beforan (approved):

PR #131, opened 2021-03-04
AB#31856: Added DisplayOnDirectory Flag To OntologyTerm

Flemingjp:
[AB#31856](https://dev.azure.com/UniversityOfNottingham/bd2de5ba-2a41-4bdc-a444-61e6b706132a/_workitems/edit/31856)

- Added DisplayOnDirectory flag To OntologyTerm
- Amended Directory service methods to filter via the DisplayOnDirectory flag


src/Directory/Biobanks.Web/Controllers/SearchController.cs
```diff
@@ -217,21 +218,32 @@ public async Task<JsonpResult> SearchOntologyTerms(string searchDocumentType, st
...
-                   MatchingOtherTerms = x.MatchingOtherTerms,
-                   NonMatchingOtherTerms = x.NonMatchingOtherTerms
-               }
-            )
-            .ToList();
+            var searchOntologyTerms = _searchProvider.ListOntologyTerms(type, wildcard);
+            var directoryOntologyTerms = await _biobankReadService.ListOntologyTermsAsync();
+
+            // Join Ontology Terms In Search and Directory Based On Ontology Term Value
+            var model = directoryOntologyTerms.Join(searchOntologyTerms,
```
Flemingjp, line 225: I moved this out from the BiobankReadService as it was more business logic that joined the results from two other services. The method was returning a DTO which didn't fit in with the BiobankReadService being a layer to access entities.

As a service method it was returning a temporary DTO model - moving into the controller dropped the need for the DTO as the result could be mapped directly to the controllers model.

blargly (approved):

beforan (approved):

PR #132, opened 2021-03-04
Added migration for PublicationAnnotations join table

JosephLavagna:
- Added migration to create PublicationAnnotations Join table after it was deleted when src/Data proj was moved to EF Core 5.
- PublicationAnnotations table will need to be deleted on Test and Prod before merge or may cause migration issues.
- PublicationAnnotation entity in Publications solution has been changed to match newly created join table.


Flemingjp (approved):

beforan (changes requested):
just blocking the merge while I check some stuff out. will review with comments after my investigations ;)

src/Data/BiobanksDbContext.cs
```diff
@@ -106,6 +106,10 @@ protected override void OnModelCreating(ModelBuilder model)
                 .HasMany(x => x.MaterialTypes)
                 .WithMany(y => y.MaterialTypeGroups);
 
+            model.Entity<Publication>()
```
beforan, line 109: I don't think this fluent config is necessary, but please confirm.

src/Data/Migrations/20210304151621_PublicationAnnotationsJoinTbl.cs
```diff
@@ -0,0 +1,47 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Biobanks.Data.Migrations
+{
+    public partial class PublicationAnnotationsJoinTbl : Migration
```
beforan, line 5: Remove this migration. A Join table already exists for `Publications` and `Annotations` in the `Initial` migration.

Test to confirm it works as needed - any configuration changes for the join entity should be done in fluent config as per these docs:

https://docs.microsoft.com/en-us/ef/core/modeling/relationships?tabs=fluent-api%2Cfluent-api-simple-key%2Csimple-key#join-entity-type-configuration

src/Directory/Data/BiobanksDbContext.cs
```diff
@@ -138,6 +138,16 @@ protected override void OnModelCreating(DbModelBuilder modelBuilder)
                     gt.ToTable("MaterialTypeMaterialTypeGroup");
                 });
 
+            modelBuilder.Entity<Publication>()
```
beforan, line 141: This config will need amending to match whatever the .NET5 context ends up doing with the join table ;)

src/Publications/Publications/PublicationDbContext.cs
```diff
@@ -25,15 +25,15 @@ public class PublicationDbContext : DbContext
         protected override void OnModelCreating(ModelBuilder modelBuilder)
         {
             modelBuilder.Entity<PublicationAnnotation>()
-                .HasKey(bc => new { bc.Publication_Id, bc.Annotation_Id });
+                .HasKey(bc => new { bc.PublicationsId, bc.AnnotationsId });
```
beforan, line 28: I notice that this PR doesn't update the Core 3.1 `LegacyData` context?

Is this a good time to move this function off of its own separate context and instead onto the `LegacyData` one. That way, we have fewer contexts and we move towards all .NET Core 3.1 code using a single context (like they will use a single .NET5/6/whatever context one day)

beforan (changes requested):

src/Publications/Publications/PublicationDbContext.cs
beforan: `LegacyData` is no more. Carry on.

PR #138, opened 2021-03-09
OntologyTerm Flagged As Visible When Written By Directory

Flemingjp:
## Overview
The OntologyTerm DisplayOnDirectory flag is defaulted as `false` by EF Core. Therefore OntologyTerms written by the Directory should be flagged as visible.

## Azure Boards
- `TODO`


src/Directory/Services/BiobankWriteService.cs
```diff
@@ -707,6 +709,8 @@ public async Task<OntologyTerm> UpdateOntologyTermAsync(OntologyTerm ontologyTer
 
         public async Task<OntologyTerm> AddOntologyTermAsync(OntologyTerm ontologyTerm)
         {
+            ontologyTerm.DisplayOnDirectory = true;
```
beforan, line 712: This is fine for now, but better to parameterise the display value, with a default of `false`, as it's conceivable the aggregator might use these service methods to insert terms from the Submission API data?

src/Directory/Services/BiobankWriteService.cs
```diff
@@ -697,6 +697,8 @@ public async Task DeleteOntologyTermAsync(OntologyTerm ontologyTerm)
 
         public async Task<OntologyTerm> UpdateOntologyTermAsync(OntologyTerm ontologyTerm)
         {
+            ontologyTerm.DisplayOnDirectory = true;
```
beforan, line 700: Not sure if this is dangerous at this time... Probably fine, pending the UI changes to make it explicit

PR #158, opened 2021-03-22
AB#32824: Removal of ExtractionProcedure and MaterialType Changes

Flemingjp:
## Overview
- Removed ExtractionProcedure entity
- Replaced ExtractionProcedure field in MaterialDetails with Many-Many OntologyTerms
- Updated MaterialType and MaterialTypeGroups seed data to capture which MaterialTypes are Extracted Samples
- Replaced `MaterialDetail` composite primary key with single `int` ID column
- Applied unique constraint on `MaterialDetail` such that the previous constraint by the old primary key still applies, with the ability of allowing null fields
- Removed `IsDefault` field from `PreservationType` as there is no need for a default field for no value - as this can be represented as `null`
- Cleaned up service code to remove reference to any default values
- Split the migrations into multiple steps - this will make migrating existing environments more manageable


src/Directory/Services/BiobankWriteService.cs
```diff
@@ -303,15 +303,15 @@ public async Task AddSampleSetAsync(CollectionSampleSet sampleSet)
         {
             // Assign default values to Material Details
             var defaultPreservationType = await _biobankReadService.GetDefaultPreservationTypeAsync();
-            var defaultExtractionProcedure = await _biobankReadService.GetDefaultExtractionProcedureAsync();
+            //var defaultExtractionProcedure = await _biobankReadService.GetDefaultExtractionProcedureAsync();
```
Flemingjp, line 307: With the changes what is our default OntologyTerm?

Flemingjp: No longer required - changing composite PK to a unique index means that null is a valid value - and implies no value, which is what the default value was doing previously

blargly (approved):

beforan (approved):

PR #159, opened 2021-03-23
AB#32827: Validation of Extraction Procedure Against Submitting Material Type

Flemingjp:
## Overview
- Updates the `ExtractionProcedure` validation in the `SampleValidationService`
    - Now validates the `ExtractionProcedure` against the submitted `MaterialType` to ensure that the provided `ExtractionProcedure` is an applicable value for the provided `MaterialType`


src/Submissions/Core/Services/SampleValidationService.cs
```diff
@@ -150,23 +151,32 @@ private async Task<StagedSample> ValidateSampleContent(SampleDto dto, StagedSamp
...
+            
+            // Invalid MaterialType or ExtractionProcedure
+            if (mt == null || ep == null)
+            {
+                throw new ValidationException(
+                    new ValidationResult(
+                        ValidationErrors.SnomedExtractionProcedure(dto.ExtractionProcedure, dto.ExtractionProcedureOntologyField, dto.Barcode, dto.IndividualReferenceId),
+                        new List<string> { nameof(dto.ExtractionProcedure) }),
+                    null, null);
+            }
```
Flemingjp, line 165: Should we be throwing an error here - thinking more about the MaterialType, as it is technically validated prior

beforan: we still need to throw for the ep though, right? so yes i think the throw is valid.

Whether the mt null check should be a condition also is debatable, but it's not exactly long and verbose and it's technically safer.

Just because we do call these methods in concert, doesn't mean this method has any dependency on any other, but it does require mt to not be null ¯\_(ツ)_/¯

beforan (approved):

blargly (approved):

PR #165, opened 2021-03-25
AB#32512: Age Range Lower and Upper Bound Seed Data

JosephLavagna:
- Added new lower and upper bound data to seed data .json


Flemingjp (approved):

sample-seed-data/AgeRange.json
```diff
@@ -2,42 +2,57 @@
     {
         "id": "1",
         "value": "Neonate (0 - 1 month)",
-        "sortOrder": "2"
+        "sortOrder": "2",
+        "lowerBound": "0M",
```
blargly, line 6: As these are durations, shouldn't they begin with a P as a designator?
https://en.wikipedia.org/wiki/ISO_8601#Durations

blargly (changes requested):

sample-seed-data/AgeRange.json
```diff
@@ -2,42 +2,57 @@
...
+        "sortOrder": "8",
+        "lowerBound": "40Y"
     },
     {
         "id": "8",
         "value": "Fetus",
-        "sortOrder": "1"
+        "sortOrder": "1",
+        "lowerBound": "-1Y",
+        "upperBound": "0"
```
beforan, line 55: are we intending to support a naked `0` value? or should it still be a valid duration (in which case `P0D` or similar would be better`?

> Date and time elements including their designator may be omitted if their value is zero, and lower-order elements may also be omitted for reduced precision. For example, "P23DT23H" and "P4Y" are both acceptable duration representations. However, at least one element must be present, thus "P" is not a valid representation for a duration of 0 seconds. "PT0S" or "P0D", however, are both valid and represent the same duration.

sample-seed-data/AgeRange.json
beforan: good catch!

Flemingjp (approved):

sample-seed-data/AgeRange.json
```diff
@@ -2,42 +2,57 @@
...
-        "sortOrder": "8"
+        "sortOrder": "8",
+        "lowerBound": "P40Y"
     },
     {
         "id": "8",
         "value": "Fetus",
-        "sortOrder": "1"
+        "sortOrder": "1",
+        "lowerBound": "P-1Y",
```
beforan, line 54: aha, this one is an oddity.

Some details covered in [this issue](https://github.com/moment/moment/issues/2408), but the upshot is:
- ISO 8601 never actually specifically allows negative durations
- .NET's `XmlConvert` (which is likely to be our implementation) uses the related `xs:duration` XML standard which does allow the WHOLE duration to be negative. i.e. `-P1D` not `P-1D`.
- so we should use that approach for negative :)

beforan (changes requested):

Flemingjp (approved):

beforan (approved):

blargly (approved):

PR #166, opened 2021-03-25
AB#31370: Prepopulate country on county choice for creating/editing sample resource page

bradthompson995:
## Overview

Adding functionality so that when creating or editing a new biobank the user can select a county option which will prepopulate the country field with the correct country. The county values are now also grouped by country.


src/Directory/Biobanks.Web/ApiControllers/CountyController.cs
```diff
@@ -152,5 +152,17 @@ public async Task<IHttpActionResult> Delete(int id)
                 name = model.Value,
             });
         }
+
+        [HttpGet]
+        [AllowAnonymous]
+        [Route("{id}/GetCountryofCounty")]
```
beforan, line 158: this is a horrible endpoint for what is at least ostensibly a REST API...

can we just call it `country`? since we know we're getting for a county because of the rest of the route.

the whole route then would be `/api/county/{id}/country` where it would return the country (id only apparently?)

that at least looks like you're requesting a property that belongs to the original resource (`county/{id}`)

Method name could also be simplified to just `GetCountryId()`? since we're in the `CountyController`, the rest is again implied.

beforan (changes requested):

src/Directory/Biobanks.Web/ApiControllers/CountyController.cs
bradthompson995: Agreed, I've changed to the names you suggested.

beforan (approved):

PR #172, opened 2021-04-07
AB#32511: Age ranges UI changes

JosephLavagna:
- Added lower and upper bounds to age ranges
- Stores them in db ISO duration format - e.g 6 months: P6M
- Displayed in UI in 'plain text' format not ISO duration
- Checks if upper bound is higher than lower bound


beforan (changes requested):

beforan (approved):

Flemingjp (dismissed):

src/Directory/Services/BiobankReadService.cs
```diff
@@ -1096,9 +1096,14 @@ public async Task<bool> IsAgeRangeInUse(int id)
             => (await GetAgeRangeUsageCount(id)) > 0;
 
         public async Task<bool> ValidAgeRangeAsync(string ageRangeDescription)
-        {
-            return (await _ageRangeRepository.ListAsync(false, x => x.Value == ageRangeDescription)).Any();
-        }
+            => (await _ageRangeRepository.ListAsync(false, x => x.Value == ageRangeDescription)).Any();
+        
+
+        public async Task<bool> ValidAgeRangeAsync(int ageRangeId, string ageRangeDescription)
```
beforan, line 1102: I think this is a misleading method name. Really it checks `IsAgeRangeDescriptionInUse` I think?

The `InUse` bit is key, otherwise checking for a non matching `Id` is quite confusing!

src/Directory/Biobanks.Web/ApiControllers/AgeRangeController.cs
```diff
@@ -94,6 +154,57 @@ public async Task<IHttpActionResult> Put(int id, AgeRangeModel model)
...
+                    if (!int.TryParse(model.UpperBound, out _))
+                    {
+                        ModelState.AddModelError("AgeRange", "Upper Bound value must be a valid number.");
+                    }
+                }
+
+            }
+            if (string.IsNullOrEmpty(model.LowerDuration) && string.IsNullOrEmpty(model.UpperDuration))
+            {
+                ModelState.AddModelError("AgeRange", "Both Upper and Lower Bounds must not be null.");
```
beforan, line 177: What about the default entry, where both were null to begin with?

beforan: (and cannot be changed in the UI)

beforan (changes requested):

src/Directory/Services/BiobankReadService.cs
JosephLavagna: Wasn't too sure about this - went off other similar methods i.e - 
`        public async Task<bool> ValidConsentRestrictionDescriptionAsync(int consentId, string consentDescription)
            => (await _collectionConsentRestrictionRepository.ListAsync(
                false,
                x => x.Value == consentDescription &&
                     x.Id != consentId)).Any();`

beforan (approved):

PR #175, opened 2021-04-08
AB#32459: Rename CollectionSampleSet -> SampleSet

thomasread99:
## Overview

Renamed the CollectionSampleSet table to SampleSet and renamed any references to it


src/Data/Migrations/20210408084227_RenameCollectionSampleSet.cs
```diff
@@ -0,0 +1,182 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Biobanks.Data.Migrations
+{
+    public partial class RenameCollectionSampleSet : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DropForeignKey(
```
beforan, line 9: This migration is pretty scary, which is a shame as the model looks like hwat we want.

Can you try using `RenameTable()` manually instead?

https://stackoverflow.com/a/50886181

beforan (changes requested):

beforan (approved):

Flemingjp (approved):

PR #177, opened 2021-04-08
AB#32943: Publicly display the version

thomasread99:
## Overview

Displays the version at /version in plain text

## Notes

Refactored some of the code to eliminate any duplicate code. New VersionHelper in Utilities contains code that was previously in HtmlHelpers\VersionHelper. This HtmlHelper now references the new VersionHelper.

### DIFF
--- a/src/Directory/Biobanks.Web/HtmlHelpers/VersionHelper.cs
+++ b/src/Directory/Biobanks.Web/HtmlHelpers/VersionHelper.cs
@@ -14,13 +14,8 @@ namespace Biobanks.Web.HtmlHelpers
 
         public static string InformationalVersion(this HtmlHelper helper)
         {
-            //InformationalVersion can contain strings, not just numbers
-            //so is useful for Semantic Versioning http://semver.org            
-
-            var attr = Utilities.VersionHelper.GetVersionNumber();            
-
             //return the assembly version if no informational version
-            return attr ?? AssemblyVersion(helper);
+            return Utilities.VersionHelper.GetInformationalVersion() ?? AssemblyVersion(helper);
         }        
     }
 }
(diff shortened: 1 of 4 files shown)


src/Directory/Biobanks.Web/HtmlHelpers/VersionHelper.cs
```diff
@@ -9,28 +9,18 @@ public static class VersionHelper
...
             //InformationalVersion can contain strings, not just numbers
-            //so is useful for Semantic Versioning http://semver.org
+            //so is useful for Semantic Versioning http://semver.org            
 
-            //http://stackoverflow.com/questions/7770068/get-the-net-assemblys-assemblyinformationalversion-value
-            var attr = Attribute.GetCustomAttribute(
-                    GetAspNetAssembly(),
-                    typeof(AssemblyInformationalVersionAttribute))
-                as AssemblyInformationalVersionAttribute;
+            var attr = Utilities.VersionHelper.GetVersionNumber();
```
beforan, line 20: honestly I think I would inline this. Especially since the `attr` variable name no longer makes sense.

src/Directory/Biobanks.Web/Utilities/VersionHelper.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Web;
+
+namespace Biobanks.Web.Utilities
+{
+    public class VersionHelper
+    {
+        public static string GetVersionNumber()
```
beforan, line 11: can we call this `GetInformationalVersion` since that is technically what it does

beforan (changes requested):

beforan (approved):

JosephLavagna (approved):

PR #181, opened 2021-04-09
AB#33562: Updated DataSeeder To Support MaterialType ExtractionProcedures

Flemingjp:
## Overview

- Added support for seeding MaterialTypes with ExtractionProcedures
- Updated `MaterialType.json` to include ExtractionProcedures for `Tissue specimen`
- Updated `OntologyTerm.json` to fix IDs that were represented in scientific notation


sample-seed-data/MaterialType.json
```diff
@@ -85,6 +85,14 @@
             {
                 "value": "Extracted Sample"
             }
+        ],
+        "extractionProcedures": [
+            {
+                "value": "Surgical procedure (procedure)"
```
beforan, line 91: would it be better to use the Ontology Term ID here (SNOMED Code currently)?

Flemingjp: Updated to use IDs

beforan (approved):

JosephLavagna (approved):

PR #191, opened 2021-04-20
AB#29896: Aggregator

Flemingjp:
First implementation of the Aggregator

- Finds all samples flagged as 'Dirty'
- Deletes those samples also flagged as 'Deleted'
- Aggregates Samples into Collections
  - Missing rules
    - [x] Mapping To AccessCondition
    - [x] Mapping To CollectionType
    - [x] Mapping To OntologyTerm
    - [x] ~~Mapping To Description~~
- Aggregates Collection samples into SampleSets
- Aggregates SampleSet samples into MaterialDetails
  - Missing rules
    - [x] Mapping To MacroscopicAssessment   
- Flags aggregated samples as 'Clean'

## Azure Boards
- [AB#29896](https://dev.azure.com/UniversityOfNottingham/bd2de5ba-2a41-4bdc-a444-61e6b706132a/_workitems/edit/29896) - Generating collections from DB
- [AB#32469](https://dev.azure.com/UniversityOfNottingham/bd2de5ba-2a41-4bdc-a444-61e6b706132a/_workitems/edit/32469) - Populate collection with sample sets
- [AB#33015](https://dev.azure.com/UniversityOfNottingham/bd2de5ba-2a41-4bdc-a444-61e6b706132a/_workitems/edit/33015) - Populate SampleSets with MaterialDetails


src/Aggregator/AzFunctions/Functions/AggregatorFunction.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            _aggregationService = aggregationService;
+        }
+
+        [Function("Aggregator")]
+        public async Task Run([TimerTrigger("*/5 * * * * *")] TimerInfo timer)
+        {
+            var dirtySamples = await _aggregationService.ListDirtySamplesAsync();
+
+            if (dirtySamples.Any())
```
beforan, line 23: This sort of structural logic isn't Azure Functions specific, is it?

It should be further abstracted, so it could be called from a console app, or aws lambda, or whatever.

The platform specific bit really should be as thin as possible, while the common bit should be as much as possible to ensure standardised behaviour across platforms

beforan (changes requested):

src/Aggregator/AzFunctions/Functions/AggregatorFunction.cs
Flemingjp: Yeah was considering the best way to further abstract this. It should be straight-forward as the function doesn't require any parameters passed to it. Just not sure what the best way to abstract this block of code - something equivalent to Java's `Task`?

Flemingjp: Solution was to move the logic to `AggregationTask` which is just a class with a `Run` function. Nothing special but seemed the simplest way to abstract while remaining clear

src/Aggregator/Core/Services/AggregationService.cs
```diff
@@ -0,0 +1,144 @@
...
+            };
+        }
+
+        public string GenerateCollectionName(LiveSample sample)
+            => string.IsNullOrEmpty(sample.CollectionName)
+                ? $"{sample.SampleContent.Value}"
+                : $"{sample.CollectionName} ({sample.SampleContent.Value})";
+
+        public async Task DeleteMaterialDetailsBySampleSetId(int id)
+            => await _db.MaterialDetails.Where(x => x.SampleSetId == id).DeleteAsync();
```
Flemingjp, line 142: Should be in a MaterialDetails service on its own?

src/Aggregator/Core/Services/CollectionService.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            _db.Update(collection);
+            await _db.SaveChangesAsync();
+        }
+
+        public async Task DeleteCollectionAsync(Collection collection)
+        {
+            await _db.Collections.Where(x => x.CollectionId == collection.CollectionId).DeleteAsync();
+            await _db.SaveChangesAsync();
+        }
```
Flemingjp, line 44: Change to DeleteById method? Also SaveChangesAsync isnt needed with DeleteAsync(), so will be removed

src/Data/BiobanksDbContext.cs
```diff
@@ -103,6 +103,12 @@ public class BiobanksDbContext : DbContext
 
         protected override void OnModelCreating(ModelBuilder model)
         {
+            // Cascade Delete
+            model.Entity<Collection>()
+                .HasMany(p => p.SampleSets)
+                .WithOne(c => c.Collection)
+                .OnDelete(DeleteBehavior.Cascade);
+
```
Flemingjp, line 111: No longer needed with manual cascade delete implemented?

src/Aggregator/Core/AggregationTask.cs
```diff
@@ -0,0 +1,120 @@
...
+                    // Write Collection To DB
+                    if (collection.CollectionId == default)
+                    {
+                        await _collectionService.AddCollectionAsync(collection);
+                    }
+                    else
+                    {
+                        foreach (var ss in oldSampleSets.Distinct())
+                        {
+                            await _aggregationService.DeleteMaterialDetailsBySampleSetId(ss);
```
Flemingjp, line 98: Thoughts on a `_materialDetailsService.DeleteBySampleSetIds(IEnumerable<int> ids)` method instead?

src/Aggregator/Core/AggregationTask.cs
```diff
@@ -0,0 +1,120 @@
...
+                        }
+
+                        await _collectionService.UpdateCollectionAsync(collection);
+
+                        // Remove old sampleSets from db if present
+                        foreach (var ss in oldSampleSets.Distinct())
+                        {
+                            await _sampleService.DeleteSampleSetById(ss);
```
Flemingjp, line 106: Same idea on multiple id delete service method

src/Aggregator/Core/AggregationTask.cs
beforan: not really fussed, all it's doing is moving where the foreach loop is, and since this is already shared code, not implementation specific, i'm not too fussed.

src/Aggregator/Core/AggregationTask.cs
beforan: same response :P

src/Data/BiobanksDbContext.cs
beforan: seems legit. try removing it and see if it requires a migration though...

src/Aggregator/Core/Services/CollectionService.cs
beforan: hold the 📞 why isn't SaveChanges required after Delete?

src/Aggregator/Core/Services/AggregationService.cs
beforan: arguably, but probably not if it's gonna be a service with one one line method?

src/Aggregator/AzFunctions/Startup.cs
```diff
@@ -8,6 +13,23 @@ class Startup
     {
         internal static void ConfigureServices(HostBuilderContext context, IServiceCollection services)
         {
+            // Register DbContext
+            services.AddDbContext<BiobanksDbContext>(options => options
+                .EnableSensitiveDataLogging()
```
beforan, line 18: what's this?

src/Aggregator/AzFunctions/Startup.cs
```diff
@@ -8,6 +13,23 @@ class Startup
...
+                .UseSqlServer(
+                    context.Configuration.GetConnectionString("Default")
+                ),
+                ServiceLifetime.Transient
```
beforan, line 22: do we need to specify this?

src/Data/BiobanksDbContext.cs
Flemingjp: Checked - no migration needed

src/Aggregator/AzFunctions/Startup.cs
Flemingjp: Not sure - was taken from [https://github.com/biobankinguk/biobankinguk/blob/0e4bc7e94353f3bb5996902a14c5fdaaa26c53b4/src/Submissions/AzFunctions/Startup.cs#L48](Submission AzFunction)

src/Aggregator/AzFunctions/Startup.cs
Flemingjp: Same as comment below

src/Aggregator/Core/Services/CollectionService.cs
Flemingjp: `DeleteAsync` is from `Z.EntityFramework.Plus` and works without `SaveChangesAsync()`. A good existing example is in `SampleService`

https://github.com/biobankinguk/biobankinguk/blob/821a96c01732f80fbea73c1cc71e23ef74079466/src/Aggregator/Core/Services/SampleService.cs#L41-L42

beforan: noice

src/Aggregator/AzFunctions/Startup.cs
beforan: can now confirm we don't need this, it's lineage dates back to the old submissions webjob.

src/Aggregator/AzFunctions/Startup.cs
beforan: can now confirm we don't need this, it's lineage dates back to the old submissions webjob.

beforan (approved):

src/Aggregator/Core/Services/Contracts/ICollectionService.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    public interface ICollectionService
+    {
+        Task<Collection> GetCollectionAsync(int organisationId, string collectionName);
+
+        Task UpdateCollectionAsync(Collection collection);
+        
+        Task AddCollectionAsync(Collection collection);
+
+        Task DeleteCollectionAsync(Collection collection);
```
beforan, line 15: remove `Async` suffix pls.

beforan (changes requested):

beforan (approved):

blargly (approved):

PR #196, opened 2021-04-23
AB#32594: Submission live function

blargly:
The original commit process relied up on Hangfire. This has now been decoupled somewhat, though the conditional switch between hangfire and the Azure Queue needs to be implemented (to be done in a later PBI).
Commit Service is now in the core class library, and is called by the new Submission Commit function.

### DIFF
--- a/src/Submissions/Api/Services/Contracts/IBackgroundJobEnqueueingService.cs
+++ b/src/Submissions/Api/Services/Contracts/IBackgroundJobEnqueueingService.cs
@@ -1,0 +1,19 @@
+﻿
+using System.Threading.Tasks;
+
+namespace Biobanks.Submissions.Api.Services.Contracts
+{
+    /// <summary>
+    /// Wrapper service for handling
+    /// </summary>
+    public interface IBackgroundJobEnqueueingService
+    {
+        /// <summary>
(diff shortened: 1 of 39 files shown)


src/Submissions/Api/Services/Contracts/IBackgroundJobEnqueueingService.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// </summary>
+    public interface IBackgroundJobEnqueueingService
+    {
+        /// <summary>
+        /// Confirms that the given Biobank is ready to commit its data to the live tables
+        /// </summary>
+        /// <param name="biobankId">Identifier for the Biobank</param>
+        /// <param name="replace">Confirms if all the current data in the live table should be deleted prior this commit</param>
+        /// <returns></returns>
+        Task QueueCommittedData(int biobankId, bool replace);
```
beforan, line 17: I don't like this name. It sounds like some data that has already been committed is being queued? Which is totally not what this does.

It's more like your queueing the request to commit some staged data

beforan (changes requested):

beforan (approved):

Flemingjp (approved):

PR #202, opened 2021-04-27
AB#32394: Generate password reset links in Biobanks area

thomasread99:
## Overview

Added functionality to the Biobanks area so that a user can generate a password reset link for other members of that Biobank.

## Notes

Upon click of the link, a modal is displayed from which you can copy the reset link to the clipboard, or view it in the browser. ADAC controller handles the link generation and displaying the modal, and the BiobankReadService handles the reset token logic.


src/Directory/Biobanks.Web/Scripts/ADAC/adac-biobanks.js
```diff
@@ -136,4 +162,9 @@ $(function () {
         );
     });
 
-})
\ No newline at end of file
+})
+
+function copyToClipboard(text) {
+    navigator.clipboard.writeText(text);    
+    alert("Copied");
```
Flemingjp, line 169: Remove the alert? Assuming for debugging at the time

Flemingjp (changes requested):

src/Directory/Biobanks.Web/Scripts/ADAC/adac-biobanks.js
thomasread99: I put the alert there so that the user knows the text has been successfully copied to their clipboard, else there is no visual feedback. Is that worth having or should I remove it?

blargly: Alert brings up a modal though right? If you want feedback that something has been copied, you want some text to appear, but not to interrupt the flow of the user. Something like a JS Toast notification would be better in this case.

thomasread99: image
This is how the new toast looks (it fades in and disappears after a second)

Flemingjp: Ah makes sense - the toast looks like a nice solution

Flemingjp (approved):

blargly (dismissed):

JosephLavagna (approved):

Flemingjp (approved):

JosephLavagna (approved):

PR #204, opened 2021-04-27
AB#34020: Content Page Entity and DbSet

thomasread99:
## Overview

Created the Content Page entity and DbSet, as well as the associated migration

### DIFF
--- a/src/Entities/Data/ContentPage.cs
+++ b/src/Entities/Data/ContentPage.cs
@@ -1,0 +1,20 @@
+﻿using Biobanks.Entities.Shared.ReferenceData;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using Biobanks.Entities.Data.ReferenceData;
+using System;
+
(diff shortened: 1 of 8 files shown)


src/Entities/Data/ContentPage.cs
```diff
@@ -0,0 +1,20 @@
...
+    public class ContentPage
+    {
+        [Key]
+        public int Id { get; set; }
+        public string RouteSlug { get; set; }
+        public string Title { get; set; }
+        public string Body { get; set; }
+        public DateTime LastUpdated { get; set; }
+        public bool IsEnabled { get; set; }
+
```
Flemingjp, line 18: Not sure if this was captured in the PBI - but some of these fields probably want a maximum length - for example RouteSlug and Title

thomasread99: What would be an ideal max length for them?

Flemingjp (approved):

blargly (approved):

PR #205, opened 2021-04-27
AB#34250: Removed HTAStatus Entity

Flemingjp:
Removes HTAStatus Entity and all references to it. In preparation for the Aggregator


JosephLavagna (approved):

blargly (approved):

Flemingjp:
Awaiting #206 + Migration

elastic-search/directory index setup/collections.json
```diff
@@ -63,9 +63,6 @@
           }
         }
       },
-      "hta": {
```
beforan, line 66: Can we add a labeler config to label `Contains: Search Index` or something, to indicate index structure changes on this PR?

It should be for the path in `./elastic-search/directory index setup/*` I think?

beforan (changes requested):

JosephLavagna (dismissed):

beforan (approved):

beforan (changes requested):
labeler won't work how we think

.github/workflows/label.yml
```diff
@@ -6,7 +6,9 @@
 # https://github.com/actions/labeler
 
 name: Labeler
-on: [pull_request]
+on:
+  pull_request:
+    types: [ready_for_review]
```
beforan, line 11: this trigger isn't quire what we want. instead we want a condition on the job itself, so it doesn't run for `draft` PRs, even though the workflow always triggers on normal PR events

beforan (changes requested):

beforan (changes requested):

beforan (approved):

JosephLavagna (approved):

PR #210, opened 2021-04-29
AB#34564: MaterialDetails FK Constraint Bug

JosephLavagna:
- When a collection was updated and a SampleSet was deleted - a FK Constraint error was thrown - MaterialDetails had a SampleSet FK.
- This manually deletes the associated MaterialDetail prior to deleting SampleSet.

### DIFF
--- a/src/Aggregator/Core/AggregationTask.cs
+++ b/src/Aggregator/Core/AggregationTask.cs
@@ -87,6 +87,10 @@ namespace Biobanks.Aggregator.Core
                     }
                     else
                     {
+                        foreach (var ss in oldSampleSets.Distinct())
+                        {
+                            await _aggregationService.DeleteMaterialDetailsBySampleSetId(ss);
+                        }
                         await _collectionService.UpdateCollectionAsync(collection);
 
                         // Remove old sampleSets from db if present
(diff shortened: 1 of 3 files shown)


src/Aggregator/Core/AggregationTask.cs
```diff
@@ -87,6 +87,10 @@ public async Task Run()
                     }
                     else
                     {
+                        foreach (var ss in oldSampleSets.Distinct())
+                        {
+                            await _aggregationService.DeleteMaterialDetailsBySampleSetId(ss);
+                        }
                         await _collectionService.UpdateCollectionAsync(collection);
```
Flemingjp, line 94: Does this have to occur before `await _collectionService.UpdateCollectionAsync(collection)` 

Could it go in the same `foreach` loop before `await _sampleService.DeleteSampleSetById(ss)`

JosephLavagna: Yep it has to because SampleSets are actually deleted in the .UpdateCollectionAsync(collection) call. And so will throw an error then. 
The reason for the manual deletion of the SampleSets is only for the run on cold start.

Flemingjp (approved):

PR #215, opened 2021-05-04
AB#34026: CRUD Methods for ContentPages

thomasread99:
## Overview

Added the CRUD functionality to the ContentPages in a ContentPageService


src/Data/Migrations/20210427112114_CreateContentPage.cs
```diff
@@ -0,0 +1,34 @@
...
+            migrationBuilder.CreateTable(
+                name: "ContentPages",
+                columns: table => new
+                {
+                    Id = table.Column<int>(type: "int", nullable: false)
+                        .Annotation("SqlServer:Identity", "1, 1"),
+                    RouteSlug = table.Column<string>(type: "nvarchar(max)", nullable: true),
+                    Title = table.Column<string>(type: "nvarchar(max)", nullable: true),
+                    Body = table.Column<string>(type: "nvarchar(max)", nullable: true),
+                    LastUpdated = table.Column<DateTime>(type: "datetime2", nullable: false),
```
Flemingjp, line 19: Doesn't this migration already exist earlier?

src/Directory/Services/ContentPageService.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            query = includeProperties.Aggregate(query, (current, includeProperty) => current.Include(includeProperty));
+
+            if (orderBy != null)
+            {
+                return orderBy.Invoke(query).ToList();
+            }
+
+            return query.ToList();
+        }
```
Flemingjp, line 63: Is this all required here? Currently it is taken from the IGenericRepository pattern, but for example the `includeProperties` could be omitted as the ContentPage entity doesn't have an Foreign Keys - would also be tempted not to required `orderBy`?

src/Directory/Services/Contracts/IContentPageService.cs
```diff
@@ -0,0 +1,23 @@
...
+    public interface IContentPageService
+    {
+        Task CreateNewContentPage(ContentPage contentPage);
+        void UpdateContentPage(ContentPage contentPage);
+        void DeleteContentPage(ContentPage contentPage);
+        IEnumerable<ContentPage> ListContentPages(bool tracking = false,
+                                    Expression<Func<ContentPage, bool>> filter = null,
+                                    Func<IQueryable<ContentPage>, IOrderedQueryable<ContentPage>> orderBy = null,
+                                    params Expression<Func<ContentPage, object>>[] includeProperties);
+        Task<ContentPage> GetById(object id);
```
Flemingjp, line 20: `id` can be `int`

src/Directory/Services/Contracts/IContentPageService.cs
```diff
@@ -0,0 +1,23 @@
...
+        Task<ContentPage> GetBySlug(object routeSlug);
```
Flemingjp, line 21: `routeSlug` can be `string`

Flemingjp (changes requested):

src/Data/Migrations/20210427112114_CreateContentPage.cs
thomasread99: I am pretty sure the migration was created with the last task & pull request I did last week

Flemingjp: Ah yes, this issue is because of #206. If you remove this migration from your branch using `dotnet ef migrations remove` and then merge `main` into this branch, it should resolve the issue.

src/Directory/Services/ContentPageService.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            IQueryable<ContentPage> query = _db.ContentPages;
+
+            if (filter != null)
+            {
+                query = query.Where(filter);
+            } 
+
+            return query.ToList();
+        }
```
Flemingjp, line 54: This can now be condensed nicely

```c#
return filter == null
  ? query.ToList()
  : query.Where(filter).ToList();
```

Flemingjp (changes requested):

src/Data/Migrations/20210427112114_CreateContentPage.cs
thomasread99: Think that's been done now, had a few merge conflicts but hopefully it's okay!

Flemingjp (dismissed):

src/Directory/Services/ContentPageService.cs
```diff
@@ -1,19 +1,67 @@
...
+
+        public async Task<ContentPage> GetById(int id)
+        {
+            return await _db.ContentPages.FindAsync(id);
+        }
+
+        public async Task<ContentPage> GetBySlug(string routeSlug)
+        {
+            return await _db.ContentPages.FindAsync(routeSlug);
```
beforan, line 63: this isn't how `FindAsync()` works. it only looks for the primary key.

so to find by slug you'll need to do `_db.ContentPages.FirstOrDefault(x -> x.Slug == routeSlug)`

src/Directory/Services/ContentPageService.cs
```diff
@@ -1,19 +1,67 @@
...
 
+        public void UpdateContentPage(ContentPage contentPage)
+        {
+            _db.ContentPages.Attach(contentPage);
+            _db.Entry(contentPage).State = EntityState.Modified;            
         }
+
+        public void DeleteContentPage(ContentPage contentPage)
+        {
+            if (_db.Entry(contentPage).State == EntityState.Detached)
```
beforan, line 39: i don't think this detached check is terribly useful.

Also, this method needs to `SaveChanges` after removing the entity.

In practice, as well, I don't think we'll ever have the entity when we ask the service to delete; it's much mroe likely we'll have the `Id` only.

(This is a side effect of designing this feature from the bottom up; we hadn't planned out the UI when work on this service started).

So can we change this to take an id parameter, then the method gets the entity by id (`FindAsync`) and if it finds nothing, just return (that's as good as deleted) or if it finds something, then `Remove` it and `SaveChangesAsync`.

src/Directory/Services/ContentPageService.cs
```diff
@@ -1,19 +1,67 @@
...
+            {
+                _db.ContentPages.Attach(contentPage);
+            }
+            _db.ContentPages.Remove(contentPage);
+        }
+
+        public IEnumerable<ContentPage> ListContentPages(bool tracking = false,
```
beforan, line 46: get rid of the arguments here.

This should just list all pages, untracked.

I was only using the repository as an example of how NOT to do `Update()`, not that this service should try and be a repository :)

src/Directory/Services/ContentPageService.cs
```diff
@@ -1,19 +1,67 @@
...
+            _db = db;
+        }
+
+        public async Task CreateNewContentPage(ContentPage contentPage)
+        {            
+            _db.ContentPages.Add(contentPage);
+            await _db.SaveChangesAsync();
+        }
 
+        public void UpdateContentPage(ContentPage contentPage)
```
beforan, line 31: This method is basically useless.

we shouldn't need to manage tracking ourselves. It's more likely the Controller will pass either its ViewModel, or a Data Transfer Object, from which we should map its properties onto our existing Entity, then `SaveChanges`.

Actually, this is pretty hard to do without the controller side model. Unless we take params, which we could for now since ContentPage is pretty lightweight.

So switch the params to `Id, Title, Body, Slug`.

Then get the existing entity by `FindAsync()` using the Id.

If you find nothing, throw a `KeyNotFoundException`.

If you find something, set its properties to the parameters:
- Title,
- Slug,
- Body
and set `LastUpdated` to todays Date/Time in UTC: `DateTime.UtcNow` or `DateTimeOffset.UtcNow` depending what the entity uses (I can't remember).

then `SaveChangesAsync()`

src/Directory/Services/Contracts/IContentPageService.cs
```diff
@@ -1,13 +1,21 @@
...
+using System.Linq.Expressions;
 using System.Text;
 using System.Threading.Tasks;
 
 namespace Biobanks.Services.Contracts
 {
     public interface IContentPageService
     {
-        Task CreateNewContentPage(string title, string body);
+        Task CreateNewContentPage(ContentPage contentPage);
```
beforan, line 13: change these method names. it's a `ContentPageService` so `Create` is fine

src/Directory/Services/ContentPageService.cs
```diff
@@ -1,19 +1,67 @@
...
     {
-        public async Task CreateNewContentPage(string title, string body)
+        private readonly BiobanksDbContext _db;
+
+        public ContentPageService(BiobanksDbContext db)
         {
+            _db = db;
+        }
+
+        public async Task CreateNewContentPage(ContentPage contentPage)
```
beforan, line 25: it would be better for this to take a model, or parameters.

Saves the calling code knowing what the Entity looks like.

So switch the params to `Title, Body, Slug`.

the initialise a new `ContentPage` entity, and set those properties on it.
and set LastUpdated to todays Date/Time in UTC.

than add that new Entity, and `SaveChangesAsync`

beforan (changes requested):

src/Directory/Services/ContentPageService.cs
```diff
@@ -1,19 +1,91 @@
...
+        }
+
+        public async Task Create(string title, string body, string slug)
+        {
+            var page = new ContentPage();
+            page.Title = title;
+            page.Body = body;
+            page.RouteSlug = slug;
+            page.LastUpdated = DateTime.UtcNow;
+
```
Flemingjp, line 32: One nice thing you can do here

```c#
var page = new ContentPage
{
  Title = title,
  Body = body,
  RouteSlug = slug,
  LastUpdated = DateTime.UtcNow
};
```

src/Directory/Services/ContentPageService.cs
```diff
@@ -1,19 +1,91 @@
...
+            }
 
+            //_db.ContentPages.Attach(contentPage);
+            //_db.Entry(contentPage).State = EntityState.Modified;            
         }
+
+        public async Task Delete(int id)
+        {
+            var page = await _db.ContentPages.FindAsync(id);
+            if (page == null)
```
Flemingjp, line 61: We can inverse the null check, if the null condition does nothing

```c#
if (page != null)
{
  _db.ContentPages.Remove(page);
  await _db.SaveChangesAsync();
}
```

src/Directory/Services/ContentPageService.cs
```diff
@@ -1,19 +1,91 @@
...
+                throw new KeyNotFoundException();
+            }
+            else
+            {
+                page.Title = title;
+                page.Body = body;
+                page.RouteSlug = slug;
+                page.LastUpdated = DateTime.UtcNow;
+                await _db.SaveChangesAsync();
+                return;
```
Flemingjp, line 51: Can remove the return statement - it will return at the end of the method anyways

src/Directory/Services/ContentPageService.cs
```diff
@@ -1,19 +1,91 @@
...
+            _db.ContentPages.Add(page);
+            await _db.SaveChangesAsync();
+        }
+
+        public async Task Update(int id, string title, string body, string slug)
         {
+            var page = await _db.ContentPages.FindAsync(id);
+            if (page == null)
+            {
+                throw new KeyNotFoundException();
```
Flemingjp, line 42: Might be no need to throw an error here? Not sure what the convention should be @beforan

src/Directory/Services/ContentPageService.cs
```diff
@@ -1,19 +1,91 @@
...
+        public IEnumerable<ContentPage> ListContentPages()
+        {
+            IQueryable<ContentPage> query = _db.ContentPages;
+
+            return query.ToList();                        
+        }
+
+        public async Task<ContentPage> GetById(int id)
+        {
+            return await _db.ContentPages.FindAsync(id);
```
Flemingjp, line 82: More a stylistic comment - but single lines can be inlined  `=> await _db.ContentPages.FindAsync(id);`

src/Directory/Services/ContentPageService.cs
```diff
@@ -1,19 +1,91 @@
...
+        }
+
+        public async Task<ContentPage> GetBySlug(string routeSlug)
+        {
+            return await _db.ContentPages.FirstOrDefaultAsync(x => x.RouteSlug == routeSlug);
```
Flemingjp, line 87: Same here `=> await _db.ContentPages.FirstOrDefaultAsync(x => x.RouteSlug == routeSlug);`

src/Directory/Services/ContentPageService.cs
```diff
@@ -1,19 +1,91 @@
...
+            else
+            {
+                _db.ContentPages.Remove(page);
+                await _db.SaveChangesAsync();
+                return;
+            }            
+        }
+
+        public IEnumerable<ContentPage> ListContentPages()
+        {
```
Flemingjp, line 74: Change to Async method and can be condensed `=> _db.ContentPages.ToListAsync()`

Flemingjp (changes requested):

src/Directory/Services/ContentPageService.cs
thomasread99: Was not able to condense this one - I think to do with datatypes etc but let me know if there's something I have missed!

Flemingjp (changes requested):

PR #218, opened 2021-05-06
AB#34505: Support Non-Extracted Samples In The Aggregator

Flemingjp:
Added support for non-extracted samples

- Relevant non-extracted samples are added to collections created by extracted samples
- Remaining non-extracted samples are aggregated into a 'Fit and Well' collection
- Collections now cascade delete automatically
- Samples are now deleted at the very end of aggregation
   - This is to prevent premature deletion in case of an aggregation error


src/Core/Aggregator/Services/OrganisationService.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Biobanks.Aggregator.Core.Services.Contracts;
+using Biobanks.Data;
+using Biobanks.Entities.Data;
+using Biobanks.Entities.Data.ReferenceData;
+using Microsoft.EntityFrameworkCore;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Biobanks.Aggregator.Core.Services
```
beforan, line 9: There is already a Shared OrganisationService in Core. please use / augment it :)

beforan (changes requested):

src/Core/Aggregator/Constants/OntologyTerms.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Biobanks.Aggregator.Constants
+{
+    public static class OntologyTerms
+    {
+        public const string FitAndWell = "102499006";
```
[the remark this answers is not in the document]
Flemingjp: Yeah I agree environmental variables is probably the best way forward - same with the MacroscopicAssessments

src/Core/Aggregator/AggregatorOptions.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// mappings for using just the ContentMethod or a configured default value.
+        /// </summary>
+        /// <returns>The mapped MacroscopicAssessment value</returns>
+        public string MapToMacroscopicAssessment(string contentMethod, string contentId)
+        {
+            // Find Mapping - Using Fallback Values If No Direct Mapping Exists
+            var map = 
+                MacroscopicAssessmentMappings.FirstOrDefault(x => x.ContentMethod == contentMethod && x.ContentId == contentId) ??       
+                MacroscopicAssessmentMappings.FirstOrDefault(x => x.ContentMethod == contentMethod && string.IsNullOrEmpty(x.ContentId)) ??   
+                MacroscopicAssessmentMappings.First(x => string.IsNullOrEmpty(x.ContentMethod) && string.IsNullOrEmpty(x.ContentId));
```
beforan, line 32: is this guaranteed to find something? or should it be `FirstOrDefault()` for the last one too?

because your reutrn value null coalesces map, suggesting it can be null after assignment.

src/Submissions/Api/Startup.cs
```diff
@@ -78,6 +78,8 @@ public void ConfigureServices(IServiceCollection services)
             var workersConfig = Configuration.GetSection("Workers").Get<WorkersOptions>() ?? new();
             var hangfireConfig = Configuration.GetSection("Hangfire").Get<HangfireOptions>() ?? new();
 
+            var aggregatorConfig = Configuration.GetSection("Aggregator").Get<AggregatorOptions>();
```
beforan, line 81: you only need this assignent if you need to use the config inside `ConfigureServices()`.

it doesn't look like you do, so the `.Configure<T>()` call below is enough.

Honestly it's a bit annoying the others have to be in both places, but you can't use the registered service version (that `.Configure<T>()` does) until the serice provider is built, which happens after `ConfigureServices()` completes.

src/Submissions/Api/appsettings.json
```diff
@@ -23,6 +23,47 @@
     }
   },
 
+  "Aggregator": {
+    "NonExtractedOntologyTerm": "102499006",
+    "MacroscopicAssessmentMappings": [
```
beforan, line 28: is it appropriate for these to be default configurations? or should they always be environment specific.

They won't work without matching ref data, right?

beforan (changes requested):

src/Core/Aggregator/AggregatorOptions.cs
Flemingjp: This depends how we want to handle no mapping. Do we throw and log this as an error, or do we allow a null macroscopic assessment to go through? Is it valid for a MaterialDetail to have a null MacroscopicAssessment

Flemingjp: This method now return null when no valid mapping. However where it is used will still throw an error, as MacroscopicAssessment is a required field. We should figure out how we handle such errors

beforan (approved):

blargly (approved):

PR #226, opened 2021-05-11
AB#35211: Publications Job

Flemingjp:
Moves the Publications Azure Function to the new Core project and abstracts the function to a Jobs class with a single `Run()` method.

Publication Job is then scheduled via Hangfire in the Submissions API


src/Submissions/Api/Startup.cs
```diff
@@ -162,6 +165,9 @@ public void ConfigureServices(IServiceCollection services)
             //TODO Register these services if we're using hangfire
             //.AddTransient<IRejectService, RejectService>()
             //.AddTransient<ICommitService, CommitService>()
+
+            // Hangfire Recurring Jobs
+            RecurringJob.AddOrUpdate<PublicationsJob>("job-publications", x => x.Run(), Cron.Daily);
```
Flemingjp, line 170: Is this the best place to define these jobs?

beforan: i say "them all" when obvs that doesn't apply to this PR, but we know there will be other recurring jobs in future

beforan (changes requested):

src/Core/Shared/Services/BiobankReadService.cs
```diff
@@ -0,0 +1,27 @@
...
+    public class BiobankReadService : IBiobankReadService
+    {
+        private BiobanksDbContext _db;
+
+        public BiobankReadService(BiobanksDbContext db)
+        {
+            _db = db;
+        }
+
+        public async Task<IList<Organisation>> ListBiobanksAsync(string wildcard = "",
```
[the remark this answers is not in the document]
beforan: yeah we should

beforan (approved):

blargly (approved):

PR #230, opened 2021-05-12
AB#35080: Analytics Hangfire Job

Flemingjp:
Refactors the Analytics Batch job into a Worker Job that is scheduled quarterly via Hangfire in the API.


src/Core/Analytics/Services/GoogleAnalyticsReadService.cs
```diff
@@ -0,0 +1,814 @@
...
+    {
+        private readonly string _viewId;
+        private readonly string _startDate; //default: "2016-01-01"; //specified in _dateFormat
+        private readonly string _dateFormat = "yyyy-MM-dd";
+
+        private readonly BiobanksDbContext _db;
+        private readonly GoogleCredential _credentials;
+        private readonly AnalyticsReportingService _analytics;
+        private readonly ILogger<GoogleAnalyticsReadService> _logger;
+        private readonly IOrganisationService _organisationService;
```
Flemingjp, line 33: Should we be looking to move any service method that uses the DbContext here?

src/Core/Analytics/Services/GoogleAnalyticsReadService.cs
```diff
@@ -0,0 +1,814 @@
...
+                Count = totalCount - sources.Select(x => x.Count).Sum(),
+                Percentage = 100.0 - sources.Select(x => x.Percentage).Sum()
+            });
+
+            return sources.ToList();
+        }
+
+        //typically performed quatertly. Should be hit by scheduler
+        public async Task UpdateAnalyticsData()
+        {
```
Flemingjp, line 786: Should this method's content be moved into the AnalyticsJob itself - this is what the job currently invokes.

beforan: yeah can do if that makes sense, either as a private helper method or inline.

It certainly doesn't need to be in _this_ service.

src/Core/Analytics/Services/GoogleAnalyticsReadService.cs
beforan: yes 100% agree. my branch does this for the http dependency bits.

if you merge first i can do the refactor, or you can look at my branch and i'll deal with potential conflicts i guess?

i have done:
- split out the `AnayticsReportGenerator` into 3:
  - Organisation (`OrganisationReportGenerator`)
  - Directory (`DirectoryReportGenerator`)
  - Shared Data Transformation helpers (`ReportDataTransformationService`)
  - Your job work shouldn't care about this though, but these are all the methods from GA service that were only modifying data in memory
-  created an `AnalyticsService` for reading any and all data from the Directory DB's Analytics tables
  - so that's the answer to this specific question
- renamed this service `GoogleAnalyticsReportingService`
  - `ReadService` isn't really a thing we do now, and there's no `WriteService` to distinguish it from anyway
  - technically the Google API we're interacting with is called `Analytics Reporting`

beforan: branch is `analytics-report-endpoints`

beforan: i guess in the job's case, it'll mostly be adding write methods to `AnalyticsService`

Flemingjp: Sounds great - should make it a lot clearer

beforan (approved):

blargly (approved):

PR #231, opened 2021-05-13
AB#35079: Analytics reports API endpoints

beforan:
## Overview

This PR adds the old HTTP Functions for Analytics Reports to the API instead, and does some refactoring to line both the endpoints and worker jobs side of the Analytics service.

Includes Directory changes to call the new endpoints.


src/Core/Analytics/Services/Contracts/IGoogleAnalyticsReportingService.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <para>
+        /// Get a Google AnalyticsReporting DateRange from
+        /// correctly formatted string values for start and end dates.
+        /// </para>
+        /// <para>Strings should be dates in the ISO-8601 short form ("yyyy-MM-dd").</para>
+        /// </summary>
+        /// <param name="startDate"></param>
+        /// <param name="endDate"></param>
+        DateRange DateRangeFromBounds(string startDate, string endDate);
+
```
Flemingjp, line 70: Could these DateRange methods be classed as helper methods and be excluded from the service definition? It would require changing the parameters of `DownloadAllBiobankData` and `DownloadDirectoryData` and calling from within.

src/Core/Analytics/Services/GoogleAnalyticsReportingService.cs
```diff
@@ -0,0 +1,358 @@
...
+        public DateRange DateRangeFromBounds(string startDate, string endDate)
+            => new() { StartDate = startDate, EndDate = endDate };
+
+        /// <inheritdoc />
+        public (DateTimeOffset startDate, DateTimeOffset endDate) GetDateRangeBounds(DateRange dateRange)
+            => (DateTimeOffset.Parse(dateRange.StartDate), DateTimeOffset.Parse(dateRange.EndDate));
+
+        /// <inheritdoc />
+        public async Task DownloadAllBiobankData(IList<DateRange> dateRanges)
+        {
```
Flemingjp, line 93: Following here, in all current usage only a single DateRange is passed here - so could accept plain DateTimeOffsets `start` and `end` instead

beforan: yeah I've no idea why this is an array. @shakisparki ?

src/Core/Analytics/Services/Contracts/IGoogleAnalyticsReportingService.cs
beforan: yeah could do. i was already trying to confine the use of DateRange to this service as much as possible, so yeah, we shouold actually go the whole way

src/Core/Analytics/Services/GoogleAnalyticsReportingService.cs
beforan: I assume an implementation detail of the GA API library, so we can absorb it within our service methods

Flemingjp (approved):

blargly (approved):

PR #228, opened 2021-05-12
AB#35247: Remove Publications AzFunction Project

Flemingjp:
Removes old Publication AzFunction in favour of the newer PublicationsJob in Core

PR #232, opened 2021-05-13
AB#35078: Submission hangfire implementation

blargly:
Ready for review


blargly:
Still need to sort out the configuration bits which set it to use hangfire.

Also needs ISubmissionService rolling back, as I made many of the methods private not realising they're used elsewhere.  maybe it needs to be split up> Old function logic could go in its own service, though not sure its warranted

src/Core/Submissions/Services/SubmissionService.cs
```diff
@@ -0,0 +1,368 @@
...
+
+        // Automapper
+        private readonly IMapper _mapper;
+
+        // Entity Validation and Write Services
+        private readonly IDiagnosisWriteService _diagnoses;
+        private readonly ISampleWriteService _samples;
+        private readonly ITreatmentWriteService _treatments;
+        private readonly IErrorService _errors;
+
```
beforan, line 34: lots of constructor injection needed?

beforan (changes requested):
Also need to delete/modify some CI bits:

- labeler configuration
- Submissions AzFunctions build workflow
- Submissions AzFunctions artifact bits inside `main.release.yml`

you should be able to mostly bae these changes on the equivalent ones in #234 or #235 or #228

Flemingjp (commented):
Will approve once outstanding comments are resolved

Flemingjp (approved):

.github/labeler.yml
```diff
@@ -11,6 +11,9 @@
   - src/Core/Jobs/AnalyticsJob.cs
 "Area: Submissions":
   - src/Submissions/**/*
+  - src/Core/Jobs/RejectJob.cs
```
beforan, line 14: what about staging?

beforan (changes requested):

beforan (approved):

Flemingjp (approved):

PR #239, opened 2021-05-19
AB#32818: Mark API Collections as such and rules around GUI editing/viewing

shakisparki:
## Overview

Description field remains editable, everything else is read only.


src/Directory/Biobanks.Web/Controllers/BiobankController.cs
```diff
@@ -960,6 +960,28 @@ public async Task<ActionResult> EditCollection(EditCollectionModel model)
             if (biobankId == 0)
                 return RedirectToAction("Index", "Home");
 
+            if (model.FromApi)
```
Flemingjp, line 963: Is it possible to do this check without relying on the model from the frontend? If you retrieve the collection beforehand, you can check if it's from the API using the backend data - this seems more reliable.

```c#
var collection = await _biobankReadService.GetCollectionByIdAsync(model.Id);

if (collection.FromApi)
{
}
```

shakisparki: You are absolutely right

src/Directory/Biobanks.Web/Scripts/Biobank/collection-crud.js
```diff
@@ -0,0 +1,12 @@
+﻿$(function () {
+    var fromapi = $("#FromApi").val().toLowerCase();
+    
+    if (fromapi == "true") {
```
beforan, line 4: `===`, and perhaps also a comment to clarify that this is NOT a boolean value, it's a string because it comes from a form field.

I hate this code, but JS doesn't have anything better without us doing a LOT more work on preparing our viewmodels for JS usage... so for now we'll just warn people of its points of confusion.

beforan (changes requested):

Flemingjp (approved):

beforan (approved):

PR #242, opened 2021-05-20
AB#35741: Bug - Selecting a datatype for associated data without a corresponding provision time throws an exception

shakisparki:
## Overview

- Displays validation message for incomplete associated data fields
- prevents exception caused by null fields for associated data during editing

## Repro 
Go to a collection > Edit > scroll down to associated data > select a checkbox but don't select a radio button for time.

### DIFF
--- a/src/Directory/Biobanks.Web/Views/Biobank/_AssociatedData.cshtml
+++ b/src/Directory/Biobanks.Web/Views/Biobank/_AssociatedData.cshtml
@@ -7,7 +7,7 @@
 <div class="row">
     <div class="form-group">
         <div class="col-sm-offset-2 col-sm-10">
-            <table class="table table-striped">
+            <table id="Groups" class="table table-striped">
                 <thead>
                     <tr>
                         <th>Data type</th>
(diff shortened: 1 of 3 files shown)


src/Directory/Biobanks.Web/Views/Biobank/_AssociatedData.cshtml
```diff
@@ -20,13 +20,14 @@
                         <tr><td colspan="5" class="info">@Html.Raw(Model.Groups[i].Name)</td></tr>
                         for (int j = 0; j < Model.Groups[i].Types.Count(); j++)
                         {
-                          @Html.EditorFor(x => Model.Groups[i].Types[j])               
+                            @Html.EditorFor(x => Model.Groups[i].Types[j])
                         }
```
Flemingjp, line 24: Can this be replaced with a foreach loop?

Flemingjp (changes requested):

src/Directory/Biobanks.Web/Views/Biobank/_AssociatedData.cshtml
shakisparki: There is a custom editor model for that in a separate file. I would keep it that way.

Flemingjp (approved):

blargly (approved):

PR #244, opened 2021-05-24
AB#33933: Blocklist/Allowlist UI

shakisparki:
## Overview
ADAC users can manage the block/allow list

Paginated datatable with filters.
Columns: Value (domain), Actions (edit, delete)

Have two main tables. Allow first, then block. Allow them to collapse for user ease.
Add new rule button - will bring up a modal form, which has value (text) input and block/allow radio button.
Stop user from adding multiple rules against the same exact domain - give a feedback message.
Edit should only edit the value, not type.


beforan (approved):

Flemingjp (changes requested):
This will need a `Mvc.sitemap` entry?

Flemingjp (approved):

PR #256, opened 2021-06-03
AB#34164: Manually Add Publications

shakisparki:
## Overview

User could input pubmed id value, and add the returned publication to their Biobank. 

Call epmc API directly and look up the publication by the ID.

- User presses a button (in publications tab) which says "Add Publication"
- Brings up a modal with an input field for the pubmed Id
- Look at local db, see if its already associated (approved, rejected or just in the suggested list). If we have no result we then hit the API for the publication.
- Return publication information below the input and search.
- Then we also have an approve and cancel button and some text (If this is the correct publication, then press the approve button to add this publication to your Biobanks approved publications list.). Cancel just closes the modal.
- On Approve, return to index view with a Publication Added validation message, and update page.

BUG FIX: fixes bug where publication adds duplicate entry on accepting or rejecting a publication


Flemingjp (approved):

Flemingjp (commented):
Will approve once conflict resolved

src/Directory/Biobanks.Web/Controllers/BiobankController.cs
```diff
@@ -1766,23 +1766,129 @@ public async Task<JsonResult> GetPublicationsAjax()
...
-                    Authors = x.Authors,
-                    Year = x.Year,
-                    Journal = x.Journal,
-                    DOI = x.DOI,
-                    Approved = x.Accepted,
-                    Source = x.Source
+                    using var client = new HttpClient();
+                    var buildUrl = new UriBuilder(ConfigurationManager.AppSettings["EpmcApiUrl"])
+                    {
+                        Path = "/europepmc/webservices/rest/search",
```
Flemingjp, line 1794: The Publications job uses `https://www.ebi.ac.uk/europepmc/` as the base URL - where here it assumes `https://www.ebi.ac.uk/` 

We should try keep it consistent between the two to avoid confusion in the configuration

Flemingjp (changes requested):

Flemingjp (approved):

blargly (approved):

PR #257, opened 2021-06-03
AB#36366: Analytics update

blargly:
Added a new DTO for storing AnalyticsReportingKey, and converted from appsettings to this new object using IOptions.


src/Submissions/Api/appsettings.json
```diff
@@ -27,5 +27,12 @@
     "EpmcApiUrl": "https://www.ebi.ac.uk/europepmc/"
   },
 
+  "Analytics": {
```
beforan, line 30: do we need the empty configs here? do they all default to null and cause problems or something? you can fix that by initializing the properties on the Options class?

src/Core/Analytics/Dto/GoogleAnalyticsReportingKey.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Biobanks.Analytics.Dto
+{
+    public class GoogleAnalyticsReportingKey
+    {
+        public string type { get; set; }
```
Flemingjp, line 5: Worth using the `[JsonProperty("")]` annotation to keep the field names C# compliant?

src/Submissions/Api/appsettings.json
beforan: Also `GoogleAnalyticsReportingKey` being an object could be really annoying to confiure in Azure Portal, KeyVault etc... a string would be simpler if we can get it working

src/Core/Analytics/Dto/GoogleAnalyticsReportingKey.cs
blargly: Sorted

src/Submissions/Api/appsettings.json
blargly: Gone with an object as its the only option I could get working.

re the appsettings - I've left it empty simply because it's easier for users to see what properties can be populated. Can appreciate that it's not essential though.

src/Directory/Services/AnalyticsService.cs
```diff
@@ -41,20 +43,19 @@ public AnalyticsReportGenerator(IBiobankReadService biobankReadService)
             }
         }
 
-        public async Task<ProfileStatusDTO> GetProfileStatus(string biobankId)
+        private async Task<ProfileStatusDTO> GetProfileStatus(Organisation biobank)
```
Flemingjp, line 46: Just two warnings to clear up here

Flemingjp (changes requested):

Flemingjp (approved):

src/Directory/Services/AnalyticsService.cs
```diff
@@ -41,20 +43,19 @@ public AnalyticsReportGenerator(IBiobankReadService biobankReadService)
...
 
             bool missingSampleSet = false;
             ProfileStatusDTO profileStatus = new ProfileStatusDTO();
 
-            foreach (var col in bb.Collections)
+            foreach (var col in biobank.Collections)
             {
-                var ss = await _biobankReadService.GetCollectionWithSampleSetsByIdAsync(col.CollectionId);
-                if (ss.SampleSets.Count == 0)
+                var collection = _db.Collections.Include("SampleSets").SingleOrDefault(x => x.CollectionId == col.CollectionId);
```
beforan, line 57: this should still be async (use `SingleOrdefaultAsync()`), and therefore so should `GetProfileStatus()`

src/Directory/Services/AnalyticsService.cs
```diff
@@ -8,25 +8,27 @@
...
         private readonly string _apiUrl = ConfigurationManager.AppSettings["DirectoryApiUrl"] ?? "";
         private readonly string _apiClientId = ConfigurationManager.AppSettings["DirectoryApiClientId"] ?? "";
         private readonly string _apiClientSecret = ConfigurationManager.AppSettings["DirectoryApiClientSecret"] ?? "";
 
         private readonly HttpClient _client;
-        private readonly IBiobankReadService _biobankReadService;
+        private readonly BiobanksDbContext _db;
 
-        public AnalyticsReportGenerator(IBiobankReadService biobankReadService)
+        public AnalyticsService(BiobanksDbContext db)
```
beforan, line 29: really though, why bother renaming this service?

- it was already accurately named
- it gave a clearer indication of what it was used for than just "some kind of analytics stuff"
- it lines up with the equivalent analytics report generation services in Core

src/Directory/Services/AnalyticsService.cs
```diff
@@ -95,10 +96,10 @@ public async Task<ProfileStatusDTO> GetProfileStatus(string biobankId)
 
         public async Task<BiobankAnalyticReportDTO> GetBiobankReport(int Id, int year, int quarter, int period)
         {
-            var bb = await _biobankReadService.GetBiobankByIdAsync(Id);
-            var biobankId = bb.OrganisationExternalId;
+            var biobank = _db.Organisations.SingleOrDefault(x => x.OrganisationId == Id);
```
beforan, line 99: this should still be async

src/Submissions/Api/appsettings.json
beforan: we need to decide to be consistent across all our code re: empty settings in appsettings.json, basically. and becuase of defaulting in Options classes, my vote is to leave them out (but document them)

beforan (changes requested):

src/Directory/Services/AnalyticsService.cs
blargly: Is it accurate? It retrieves some data from the API for directory or biobank level reports. The "generating" of that report happened on the API side I would argue.

PR #268, opened 2021-06-14
AB#32151: Add ExtractionProcedure API Controller and Service Methods

shakisparki:
## Overview

Add an API Controller for Extraction Procedure - and create the required service methods alongside.

Note ExtractionProcedure is now a SnomedTerm and will need to be filtered from that table

Some bug fixes.


Flemingjp (approved):

src/Directory/Biobanks.Web/ApiControllers/ExtractionProcedureController.cs
```diff
@@ -0,0 +1,152 @@
...
+                .Result
+            )
+            .ToList();
+        }
+
+        [HttpDelete]
+        [Route("{id}")]
+        public async Task<IHttpActionResult> Delete(string id)
+        {
+            var model = (await _biobankReadService.ListExtractionProceduresAsync()).Where(x => x.Id == id).First();
```
Flemingjp, line 50: Better of creating a `GetExtractionProcedure` service method? That way you can do the filtering on database

Flemingjp (changes requested):

Flemingjp (dismissed):

src/Directory/Biobanks.Web/ApiControllers/DiseaseStatusController.cs
```diff
@@ -79,14 +79,14 @@ public async Task<IHttpActionResult> Put(string id, OntologyTermModel model)
             //If this description is valid, it already exists
             if (await _biobankReadService.ValidOntologyTermDescriptionAsync(id, model.Description))
             {
-                ModelState.AddModelError("Description", "That description is already in use by another disease status. Disease status descriptions must be unique.");
+                ModelState.AddModelError("Description", "That description is already in use by another disease status or extraction procedure. Disease status descriptions must be unique.");
```
beforan, line 82: I feel like this message is confusing everywhere it appears. In this particular case, the first clause mentions both ontology term types, but the second only mentions disease status, and if they're editing disease statuses mentioning extraction procedure out of the blue is just gonna confuse them.

I think something like "That description is already in use. Descriptions must be unique across all ontology terms." is less confusing but still accurate.

src/Directory/Biobanks.Web/ApiControllers/DiseaseStatusController.cs
```diff
@@ -115,10 +115,14 @@ public async Task<IHttpActionResult> Put(string id, OntologyTermModel model)
         [Route("")]
         public async Task<IHttpActionResult> Post(OntologyTermModel model)
         {
+            //if ontology term id is in use by another ontology term
+            if ((await _biobankReadService.ListOntologyTermsAsync()).Any(x => x.Id == model.OntologyTermId))
+                ModelState.AddModelError("OntologyTermId", "This Id is currently in use by another Disease status or extraction procedure. Ontology Term Ids must be unique");
```
beforan, line 120: I would word this the same as the description message, but with "ID" instead.

src/Directory/Biobanks.Web/ApiControllers/ExtractionProcedureController.cs
```diff
@@ -0,0 +1,152 @@
...
+using System.Web.Http;
+using Biobanks.Web.Models.Shared;
+using Biobanks.Web.Models.ADAC;
+using System.Collections;
+using Biobanks.Entities.Shared.ReferenceData;
+using Biobanks.Web.Filters;
+
+namespace Biobanks.Web.ApiControllers
+{
+    //[UserApiAuthorize(Roles = "ADAC")]
```
beforan, line 13: why is this commented out? is that a mistake or should it be removed?

src/Directory/Biobanks.Web/ApiControllers/ExtractionProcedureController.cs
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        [HttpPut]
+        [Route("{id}")]
+        public async Task<IHttpActionResult> Put(string id, OntologyTermModel model)
+        {
+            //If this description is valid, it already exists
+            if (await _biobankReadService.ValidOntologyTermDescriptionAsync(id, model.Description))
+            {
+                ModelState.AddModelError("Description", "That description is already in use by another disease status or extraction procedure. Extraction procedure descriptions must be unique.");
```
beforan, line 83: Simplify this / make it consistent with elsewhere

beforan (changes requested):

blargly:
No longer needed

PR #278, opened 2021-06-25
AB#35720: Diagnosis Status DisplayOnDirectory UI

Flemingjp:
- Disease Status ADAC UI
  - Added DisplayOnDirectory Column
  - Added DisplayOnDirectory Checkbox on Add/Edit Modal
  - Displays all OntologyTerms rather than those just flagged DisplayOnDirectory
  - Moved from static generated DataTable to dynamic ajax DataTable
    - New PaginatedDiseaseStatus endpoint
    - Action links are generated on the frontend instead
    - Removal of legacy data bindings where applicable
  - Removal of model on DiseaseStatus View

### DIFF
--- a/src/Directory/Services/Contracts/IBiobankReadService.cs
+++ b/src/Directory/Services/Contracts/IBiobankReadService.cs
@@ -131,17 +131,15 @@ namespace Biobanks.Services.Contracts
         Task<bool> ValidAssociatedDataProcurementTimeFrameDescriptionAsync(string procurementDescription);
 
         Task<int> CountOntologyTerms(string filter = "");
-        Task<IEnumerable<OntologyTerm>> PaginateOntologyTerms(int start, int length, string filter = "");
-        Task<IEnumerable<OntologyTerm>> ListOntologyTerms(string filter = "", bool onlyDisplayable = false);
-        Task<IEnumerable<OntologyTerm>> ListDisplayableOntologyTerms(string filter = "");
         Task<bool> ValidOntologyTermDescriptionAsync(string OntologyTermDescription);
         Task<bool> ValidOntologyTermDescriptionAsync(string ontologyTermId, string ontologyDescription);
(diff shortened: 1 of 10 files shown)


src/Directory/Services/Contracts/IBiobankReadService.cs
```diff
@@ -130,7 +130,10 @@ public interface IBiobankReadService
         Task<bool> ValidAssociatedDataProcurementTimeFrameDescriptionAsync(int procurementId, string procurementDescription);
         Task<bool> ValidAssociatedDataProcurementTimeFrameDescriptionAsync(string procurementDescription);
 
-        Task<IEnumerable<OntologyTerm>> ListOntologyTermsAsync(string wildcard = "");
+        Task<int> CountOntologyTerms(string filter = "");
+        Task<IEnumerable<OntologyTerm>> PaginateOntologyTerms(int start, int length, string filter = "");
+        Task<IEnumerable<OntologyTerm>> ListOntologyTerms(string filter = "", bool onlyDisplayable = false);
```
beforan, line 135: 1. why isn't this async?
2. it has also essentially been replaced (or the old async version has) by the newer `ListDiseaseOntologyTermsAsync()`? (or something like that) so needs to take that filtering into account, since I think it's supposed to be getting disease statuses only

src/Directory/Services/Contracts/IBiobankReadService.cs
```diff
@@ -130,7 +130,10 @@ public interface IBiobankReadService
...
+        Task<IEnumerable<OntologyTerm>> ListDisplayableOntologyTerms(string filter = "");
```
beforan, line 136: Similar to above, I think this should also be getting Disease Statuses only, and so additional filtering may be needed.

beforan (changes requested):

beforan (approved):

blargly (approved):

PR #285, opened 2021-07-06
AB#36916: Config Service

JosephLavagna:
- Take existing config services from the read/write services
- Specific methods dealing with boolean flags only. Handles the conversion from strings to bools within the method to avoid any type issues doing it manually. Instances where a boolean value is being used should make use of this method.

- Will be creating FlagController with SuperUser flag view in PBI #36740 which will make use of the ConfigService and methods created for this PR.

### DIFF
--- a/src/Directory/Biobanks.Web/Controllers/ADACController.cs
+++ b/src/Directory/Biobanks.Web/Controllers/ADACController.cs
@@ -927,7 +927,7 @@ namespace Biobanks.Web.Controllers
         public async Task<ActionResult> Analytics(int year = 0, int endQuarter = 0, int reportPeriod = 0)
         {
             //If turned off in site config
-            if (!(await _configService.GetSiteConfigStatus(ConfigKey.DisplayAnalytics)))
+            if (await _configService.GetFlagConfigValue(ConfigKey.DisplayAnalytics) == false)
                 return RedirectToAction("LockedRef");
 
             //set default options
@@ -1210,7 +1210,7 @@ namespace Biobanks.Web.Controllers
                     .Result
                 )
                 .ToList();
-            if (await _configService.GetSiteConfigStatus("site.display.preservation.percent"))
+            if (await _configService.GetFlagConfigValue("site.display.preservation.percent") == true)
             {
                 return View(new CollectionPercentagesModel()
                 {
@@ -1527,7 +1527,7 @@ namespace Biobanks.Web.Controllers
         #region RefData: County
         public async Task<ActionResult> County()
         {
-            if (await _configService.GetSiteConfigStatus("site.display.counties"))
+            if (await _configService.GetFlagConfigValue("site.display.counties") == true)
             {
                 var countries = await _biobankReadService.ListCountriesAsync();
 
--- a/src/Directory/Biobanks.Web/Controllers/ADACController.cs
+++ b/src/Directory/Biobanks.Web/Controllers/ADACController.cs
@@ -25,6 +25,7 @@ using System.Data.Entity;
 using Newtonsoft.Json;
 using System.Net.Http;
 using Microsoft.ApplicationInsights;
+using static Biobanks.Services.ConfigService;
 
 namespace Biobanks.Web.Controllers
 {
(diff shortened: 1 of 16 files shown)


src/Directory/Services/ConfigService.cs
```diff
@@ -0,0 +1,94 @@
...
+        public async Task<IEnumerable<Config>> ListSiteConfigsAsync(string wildcard = "")
+            => await _db.Configs.Where(x => x.Key.Contains(wildcard)).ToListAsync();
+
+        public async Task<Config> GetSiteConfig(string key)
+             => (await ListSiteConfigsAsync(key)).FirstOrDefault();
+
+        public async Task<string> GetSiteConfigValue(string key, string defaultValue = "")
+            => (await GetSiteConfig(key))?.Value ?? defaultValue;
+
+        public async Task<bool> GetSiteConfigStatus(string siteConfigValue)
```
beforan, line 34: I don't think this method (or its argument) is clearly named. But also, what is its purpose?

It seems to do the same job as `GetFlagConfigValue()`, but less well.

src/Directory/Services/ConfigService.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            }
+            await _db.SaveChangesAsync();
+        }
+
+        // Getter - Returns boolean value of flag, null if not flag
+        public async Task<bool?> GetFlagConfigValue(string key)
+            => (ConvertStringToBool((await GetSiteConfig(key)).Value));
+
+        public async Task<List<Config>> ListBooleanFlags(bool includeFeatures, string wildcard = "")
```
beforan, line 57: `includeFeatures` is mis-named. the functionality is actually "not features" when it's `false`, and "only features" when it's `true`. To get ALL boolean config records, you'd have to call it twice.

Not sure what the fix is. If this provides the required functionality adequately, then probably just rename the argument? Not sure what to though. `isFeatureFlag` probably, since then it's setting the expected value of the column (though for false, null is also valid... maybe a doc comment could clear that up)

beforan (changes requested):

src/Directory/Services/ConfigService.cs
JosephLavagna: This method was moved over from ReadService but yep it does seem to do the same thing as `GetFlagConfigValue()` - will remove the method and update references

src/Directory/Services/ConfigService.cs
JosephLavagna: Yep wasn't too sure on this - I could make it a nullable bool - which would allow for the three different possible scenarios: 
(1) - Get all boolean flags (no matter if feature or not) - (param = null)
(2) - Get boolean flags only where `isFeatureFlag` == true - (param = true)
(3) - Get boolean flags only where `isFeatureFlag` == false || null - (param = false)

Not ideal, and would not work with the idea of the expected value of the `isFeatureFlag` column. 
Or could separate into two methods - one which just grabs  all boolean flags and another which grabs all flags(bool isFeature).

beforan: yeah i thought about nullable bool but i thought that would be misleading in it's own way, especially since the column is also nullable (but for the column null means false!).

Arguably the actually right thing to do would be an enum that allows you to specify which of the 3 behaviours you want.
Actually, I think that's probably right. It can be in this file since it's an option exclusively for this service. 

so something like 

```csharp
enum BooleanConfigSelection {
  FeatureFlagsOnly
  ExcludeFeatureFlags
  AllBooleanConfigs
}
```

then default the the argument to AllBooleanConfigs, and then if people want only flags, or only NOT flags, they can specify.

Not wedded to those names, if you can find a clearer way of expressing them, go for it :)

beforan (approved):

src/Directory/Biobanks.Web/Controllers/ADACController.cs
```diff
@@ -25,6 +25,7 @@
 using Newtonsoft.Json;
 using System.Net.Http;
 using Microsoft.ApplicationInsights;
+using static Biobanks.Services.ConfigService;
```
shakisparki, line 28: Is this necessary?

JosephLavagna: Yep it is - when calling the  `ListBooleanFlags()` method you need to pass it a 'selection' of the enum - `BooleanConfigSelection` and this needs that reference to do so.

shakisparki (approved):

src/Directory/Biobanks.Web/Controllers/ADACController.cs
beforan: i assume it's because it's referencing the enum member of the class, in which case, yes.

PR #289, opened 2021-07-13
AB#36740: SuperUser/Flags View

JosephLavagna:
- Created new controller - FlagsController
- Created new view - FlagsConfig
- View contains two tables - boolean configs (no features) and feature configs


src/Directory/Biobanks.Web/Views/SuperUser/Flags.cshtml
```diff
@@ -0,0 +1,123 @@
+﻿@using System.Web.Optimization
```
beforan, line 1: is this view still needed if we have the general flags one?

JosephLavagna: Good spot thought I deleted it, it's not in the solution explorer in VS but can see it in file explorer on Windows - that means the reference must have been deleted from the .csproj file right?

beforan: yeah it's not in the csproj else we'd see it added to the diff in this PR :)

beforan (approved):

blargly (approved):

PR #293, opened 2021-07-20
AB#38414: Include Findings in Search Typeahead

Flemingjp:
Currently the search typeahead will only show results for Ontology Disease Terms that are found in the search index. This means the often used `Fit and well (finding)` term is excluded, as it is a Finding. 

This PR introduces findings to be included in the search typeahead

### DIFF
--- a/src/Directory/Biobanks.Web/Controllers/SearchController.cs
+++ b/src/Directory/Biobanks.Web/Controllers/SearchController.cs
@@ -245,10 +245,13 @@ namespace Biobanks.Web.Controllers
         private async Task<List<OntologyTermModel>> GetOntologyTermSearchResultsAsync(SearchDocumentType type, string wildcard)
         {
             var searchOntologyTerms = _searchProvider.ListOntologyTerms(type, wildcard);
-            var directoryOntologyTerms = await _biobankReadService.ListDiseaseOntologyTermsAsync("", onlyDisplayable: true);
+            var directoryDiseaseTerms = await _biobankReadService.ListDiseaseOntologyTermsAsync(onlyDisplayable: true);
+            var directoryFindingTerms = await _biobankReadService.ListFindings(onlyDisplayable: true);
 
             // Join Ontology Terms In Search and Directory Based On Ontology Term Value
-            var model = directoryOntologyTerms.Join(searchOntologyTerms, 
+            var model = directoryDiseaseTerms
+                .Concat(directoryFindingTerms)
+                .Join(searchOntologyTerms, 
                 outer => outer.Value.ToLower(), 
                 inner => inner.OntologyTerm, 
                 (directoryTerm, searchTerm) =>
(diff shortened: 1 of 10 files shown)


shakisparki (approved):

src/Directory/Biobanks.Web/Controllers/SearchController.cs
```diff
@@ -245,10 +245,13 @@ public async Task<JsonpResult> SearchOntologyTerms(string searchDocumentType, st
...
         {
             var searchOntologyTerms = _searchProvider.ListOntologyTerms(type, wildcard);
-            var directoryOntologyTerms = await _biobankReadService.ListDiseaseOntologyTerms("", onlyDisplayable: true);
+            var directoryDiseaseTerms = await _biobankReadService.ListDiseaseOntologyTerms(onlyDisplayable: true);
+            var directoryFindingTerms = await _biobankReadService.ListFindingOntologyTerms(onlyDisplayable: true);
 
             // Join Ontology Terms In Search and Directory Based On Ontology Term Value
-            var model = directoryOntologyTerms.Join(searchOntologyTerms, 
+            var model = directoryDiseaseTerms
+                .Concat(directoryFindingTerms)
```
beforan, line 253: ok but we could just do this in one sql call...

hate to say it, but having merged the previous changes around filtering and providing domain specific ontologyterm getters, maybe we need to refactor into one method that includes whatever tags you ask for in one go?

Will make this more efficient, simpler and easier to grok, and also remains extensible for the future.

so something like:

```csharp
List<OntologyTerm> ListOntologyTerms(List<string> includeTags = null, onlyDisplayable = (whatever the default is))
{
  var query = _db.OntologyTerms
    .Where(x => x.DisplayOnDirectory || !onlyDisplayable);
    // include the wildcard Where() clause here if filtering by wildcard
  
  if(includeTags != null)
  {  
    if(!includeTags.Any()) // get only untagged terms
      query = query.Where(x => x.SnomedTag == string.Empty || null);

    includeTags.ForEach(tag => { query = query.Where(x => x.SnomedTag == tag); }
  }
  // else - get everything regardless of tag

  return query.ToListAsync();
}
```

If the above doesn't work exactly as I think (chaining `Where`) then can use LinqKit's `PredicateBuilder` instead.

Also doc comment the `includeTags` parameter to be clear about the difference between `null` and an empty list, if you do it this way :) Open to a different and better API if you've got one though.

probably also change the tag strings from the constants they are now to an actual static constants class (this is a pattern we already use).

```csharp
public static class SnomedTags {
  public const string Disease = "Disease";
  // ...
}
```

Doing this refactor saves on the almost exact duplicating below, and provides an oportunity to escape the `GenericRepository` for these cases. ;)

beforan (changes requested):

src/Directory/Services/Constants/SnomedTags.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Biobanks.Directory.Services.Constants
+{
+    public static class SnomedTags
+    {
+        public const string Disease = "Disease";
+        public const string Finding = "Finding";
+    }
```
beforan, line 7: should there be a tag for extraction procedure?

blargly (approved):

src/Directory/Services/Constants/SnomedTags.cs
beforan: yes, but work for this is currently captured in AB#37807

beforan (approved):

PR #295, opened 2021-07-20
AB#38419: Create collection view - fit and well not showing on disease status typeahead

shakisparki:
## Overview

Go to create collection view - begin to type "Fit and Well (Finding)". This does not display in the typeahead.
Furthermore, if the entire term is entered into the field, and rest of the form is valid, an error is returned relating to the disease status field.

### DIFF
--- a/src/Directory/Services/BiobankReadService.cs
+++ b/src/Directory/Services/BiobankReadService.cs
@@ -1227,6 +1227,14 @@ namespace Biobanks.Services
                 ))
                 .Any();
 
+        public async Task<bool> ValidDiseaseFindingOntologyTermDescriptionAsync(string ontologyTermDescription)
+            => (await _ontologyTermRepository.ListAsync(
+                filter: x =>
+                    (x.SnomedTag.Value == DiseaseTag || x.SnomedTag.Value == FindingTag) &&
+                    x.Value == ontologyTermDescription &&
+                    x.DisplayOnDirectory
+                ))
+                .Any();
         #endregion
         #region RefData: Extraction Procedure
 
--- a/src/Directory/Services/BiobankReadService.cs
+++ b/src/Directory/Services/BiobankReadService.cs
@@ -1206,19 +1206,19 @@ namespace Biobanks.Services
                     .Take(length)
                     .ToListAsync();
 
-        public async Task<IEnumerable<OntologyTerm>> ListDiseaseOntologyTermsAsync(string wildcard = "", bool onlyDisplayable = false)
+        public async Task<IEnumerable<OntologyTerm>> ListDiseaseOntologyTerms(string wildcard = "", bool onlyDisplayable = false)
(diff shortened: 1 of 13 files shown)


src/Directory/Services/Contracts/IBiobankReadService.cs
```diff
@@ -138,7 +138,9 @@ public interface IBiobankReadService
         Task<bool> IsOntologyTermInUse(string id);
 
         Task<IEnumerable<OntologyTerm>> ListDiseaseOntologyTermsAsync(string wildcard = "", bool onlyDisplayable = false);
+        Task<IEnumerable<OntologyTerm>> ListDiseaseFindingOntologyTermsAsync(string wildcard = "", bool onlyDisplayable = false);
```
Flemingjp, line 141: I understand where this naming comes from, but I think it would be more concise to cut it down to something like `ListFindingOntologyTerms` - including dropping `Async`

src/Directory/Services/BiobankReadService.cs
```diff
@@ -1211,6 +1212,12 @@ public async Task<IEnumerable<OntologyTerm>> ListDiseaseOntologyTermsAsync(strin
                 x.Value.Contains(wildcard) &&
                 (x.DisplayOnDirectory || !onlyDisplayable));
 
+        public async Task<IEnumerable<OntologyTerm>> ListDiseaseFindingOntologyTermsAsync(string wildcard = "", bool onlyDisplayable = false)
```
Flemingjp, line 1215: I think it'd be much clearer to have this just return Findings, from a service method point of view. Then if both Disease and Findings are required, they are called together and the results concatenated where needed.

If this dual method is truly needed, it should be a third service method that calls the previous two.

src/Directory/Services/BiobankReadService.cs
```diff
@@ -1220,6 +1227,14 @@ public async Task<bool> ValidDiseaseOntologyTermDescriptionAsync(string ontology
                 ))
                 .Any();
 
+        public async Task<bool> ValidDiseaseFindingOntologyTermDescriptionAsync(string ontologyTermDescription)
```
Flemingjp, line 1230: Same idea here. We can validate twice

Flemingjp (changes requested):

src/Directory/Services/Contracts/IBiobankReadService.cs
blargly: This is my bad as I added this in my branch, but good point.

src/Directory/Services/BiobankReadService.cs
blargly: This is my bad, as I did this in my branch, but yes good point to seperate.

Flemingjp (approved):

blargly (approved):

PR #296, opened 2021-07-22
AB#38420: Extraction Procedure config page does not display displayinDirectory=0 entries

shakisparki:
## Overview

Adds option to turn on and off DisplayOnDirectory for extraction procedures.

Extraction procedure service methods replaced with new ontology term methods

### DIFF
--- a/src/Directory/Biobanks.Web/ApiControllers/ExtractionProcedureController.cs
+++ b/src/Directory/Biobanks.Web/ApiControllers/ExtractionProcedureController.cs
@@ -54,10 +54,10 @@ namespace Biobanks.Web.ApiControllers
         [Route("{id}")]
         public async Task<IHttpActionResult> Delete(string id)
         {
-            var model = await _biobankReadService.GetOntologyTerm(id, null, new List<string>
+            var model = await _biobankReadService.GetOntologyTerm(id, tags: new List<string>
             {
                 SnomedTags.ExtractionProcedure
-            }, false);
+            });
 
             if (await _biobankReadService.IsExtractionProcedureInUse(id))
             {
@@ -97,10 +97,10 @@ namespace Biobanks.Web.ApiControllers
             if (model.MaterialTypeIds == null || model.MaterialTypeIds.Count == 0)
                 ModelState.AddModelError("MaterialTypeIds", "Add at least one material type to the extraction procedure.");
 
-            var ontologyTerm = await _biobankReadService.GetOntologyTerm(id, null, new List<string>
+            var ontologyTerm = await _biobankReadService.GetOntologyTerm(id, tags: new List<string>
             {
                 SnomedTags.ExtractionProcedure
-            },false);
+            });
             if (await _biobankReadService.IsExtractionProcedureInUse(id))
             {
                 //Allow editing of only Other terms field if ontologyterm in use
--- a/src/Directory/Biobanks.Web/ApiControllers/ExtractionProcedureController.cs
+++ b/src/Directory/Biobanks.Web/ApiControllers/ExtractionProcedureController.cs
@@ -37,7 +37,8 @@ namespace Biobanks.Web.ApiControllers
                     Description = x.Value,
                     MaterialDetailsCount = await _biobankReadService.GetExtractionProcedureMaterialDetailsCount(x.Id),
                     OtherTerms = x.OtherTerms,
-                    MaterialTypeIds = x.MaterialTypes.Select(x => x.Id).ToList()
+                    MaterialTypeIds = x.MaterialTypes.Select(x => x.Id).ToList(),
+                    DisplayOnDirectory = x.DisplayOnDirectory
                 })
                 .Result
             )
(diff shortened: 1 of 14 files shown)


src/Directory/Services/BiobankReadService.cs
```diff
@@ -1238,24 +1238,25 @@ public async Task<bool> ValidFindingOntologyTermDescription(string ontologyTermD
...
 
-        public async Task<IEnumerable<OntologyTerm>> GetMaterialTypeExtractionProcedures(int id)
+        public async Task<IEnumerable<OntologyTerm>> GetMaterialTypeExtractionProcedures(int id, bool onlyDisplayable = false)
         => (await _materialTypeRepository.ListAsync(false, x => x.Id == id, null, x => x.ExtractionProcedures))
         .FirstOrDefault()?.ExtractionProcedures
+        .Where(x=> x.DisplayOnDirectory || !onlyDisplayable)
         .ToList();
 
-        public async Task<IEnumerable<OntologyTerm>> ListExtractionProceduresAsync(string wildcard = "")
+        public async Task<IEnumerable<OntologyTerm>> ListExtractionProceduresAsync(string wildcard = "", bool onlyDisplayable = false)
```
beforan, line 1247: most of these service methods can move to the new generalised OntologyTerms methods (where you specify which tags to include).

beforan (changes requested):

src/Directory/Services/BiobankReadService.cs
beforan: need to add Extraction Procedure tag value to the new SnomedTags constants class too

src/Directory/Biobanks.Web/ApiControllers/ExtractionProcedureController.cs
```diff
@@ -48,7 +54,10 @@ public async Task<IList> Get()
         [Route("{id}")]
         public async Task<IHttpActionResult> Delete(string id)
         {
-            var model = await _biobankReadService.GetExtractionProcedureById(id);
+            var model = await _biobankReadService.GetOntologyTerm(id, null, new List<string>
+            {
+                SnomedTags.ExtractionProcedure
+            }, false);
```
Flemingjp, line 60: Specify the parameters by name to avoid random null values, and to add clarity

```c#
var model = await _biobankReadService.GetOntologyTerm(id, onlyDisplayable: false, tags: new List<string> 
{ 
    SnomedTags.ExtractionProcedure 
});
```

src/Directory/Biobanks.Web/Controllers/BiobankController.cs
```diff
@@ -1443,7 +1445,10 @@ private async Task<AbstractCRUDSampleSetModel> PopulateAbstractCRUDSampleSetMode
                         })
                 .OrderBy(x => x.SortOrder);
 
-            model.ExtractionProcedures = (await _biobankReadService.ListExtractionProceduresAsync())
+            model.ExtractionProcedures = (await _biobankReadService.ListOntologyTerms(null, new List<string>
+                {
+                    SnomedTags.ExtractionProcedure
+                },true))
```
Flemingjp, line 1451: Same here. `onlyDisplayable` is also true by default

src/Directory/Services/BiobankReadService.cs
```diff
@@ -1220,24 +1222,12 @@ public async Task<int> GetOntologyTermCollectionCapabilityCount(string id)
 
         #region RefData: Extraction Procedure
 
-        public async Task<IEnumerable<OntologyTerm>> GetMaterialTypeExtractionProcedures(int id)
+        public async Task<IEnumerable<OntologyTerm>> GetMaterialTypeExtractionProcedures(int id, bool onlyDisplayable = false)
```
Flemingjp, line 1225: Can we standardise the `onlyDisplayable` flag default across service methods. Either all should be `true` or `false` - my preference would be to `false` such that an additional filter isnt imposed by default

Flemingjp (changes requested):

src/Directory/Biobanks.Web/ApiControllers/ExtractionProcedureController.cs
```diff
@@ -79,7 +88,7 @@ public async Task<IHttpActionResult> Delete(string id)
         public async Task<IHttpActionResult> Put(string id, ReadExtractionProcedureModel model)
         {
             //If this description is valid, it already exists
-            if (await _biobankReadService.ValidOntologyTerm(id, description: model.Description))
+            if ((await _biobankReadService.ListOntologyTerms()).Any(x=>x.Value == model.Description && x.Id != id))
```
Flemingjp, line 91: Why change to this method?

beforan (approved):

src/Directory/Biobanks.Web/ApiControllers/ExtractionProcedureController.cs
shakisparki: The use of ValidOntologyTerm in this way prevents updating an ontologyterm without changing the description.
This restricts a PUT operation for example when you want to update something else like just the material type.

Ofcourse this is potentially an issue with all other Api controllers that use the Valid...methods. However it doesn't pose a problem for most as they have one editable field which is the description. In these cases, updating the description without changing it fails but this isn't a functional problem.

Flemingjp (approved):

PR #298, opened 2021-07-27
AB#37792: Publications on prod is failing to run - looking for annotationspublication object

shakisparki:
## Overview

Local env seem to be seeding an annotationspublications object, unlike test and prod which are using publicationsannotations.

## Fix

Replaces PublicationAnnotations with AnnotationPublication object.
Sql checks ensures consistency accross all environments


src/Data/Migrations/20210727141919_FixPublicationAnnotations.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace Biobanks.Data.Migrations
+{
+    public partial class FixPublicationAnnotations : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            //Drop PublicationAnnotations and Add AnnotationPublication
+            migrationBuilder.Sql("DROP TABLE IF EXISTS dbo.PublicationAnnotations");
+            
+            migrationBuilder.Sql("IF OBJECT_ID('dbo.AnnotationPublication', 'U') IS NULL " +
```
beforan, line 12: since local migrations (a fresh environment) are correct, we don't need to store the migrations to fix existing brkoen environments in the source. we can just apply one time fixes to the broken environments, using a script derived from the SQL here.

Once those one time fixes are applied to all environments, this PR can be abandoned.

beforan (changes requested):

PR #299, opened 2021-07-28
AB#38662: Refactor Into OntologyTerm Service

Flemingjp:
Moved all OntologyTerm service methods into a dedicated `OntologyTermService`


src/Directory/Services/Contracts/IOntologyTermService.cs
```diff
@@ -0,0 +1,24 @@
+﻿using Biobanks.Entities.Shared.ReferenceData;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Biobanks.Directory.Services.Contracts
+{
+    public interface IOntologyTermService
```
beforan, line 7: Doc comments everywhere in here please.

src/Directory/Services/Contracts/IOntologyTermService.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        Task<IEnumerable<OntologyTerm>> ListOntologyTerms(string description = null, List<string> tags = null, bool onlyDisplayable = false);
```
beforan, line 9: Call it `List()`

src/Directory/Services/Contracts/IOntologyTermService.cs
```diff
@@ -0,0 +1,24 @@
...
+        Task<IEnumerable<OntologyTerm>> PaginateOntologyTerms(int start, int length, string description = null, List<string> tags = null);
```
beforan, line 10: don't need `OntologyTerm` in any of the method names; that should be the default assumption on an `OntologyTermService`

for this one, not sure if I prefer `Paginate()` or `ListPaginated()` (which I think is technically more correct)

src/Directory/Services/Contracts/IOntologyTermService.cs
```diff
@@ -0,0 +1,24 @@
...
+        
+        Task<OntologyTerm> GetOntologyTerm(string id = null, string description = null, List<string> tags = null, bool onlyDisplayable = false);
```
beforan, line 12: `Get()`

src/Directory/Services/Contracts/IOntologyTermService.cs
```diff
@@ -0,0 +1,24 @@
...
+        
+        Task<bool> ValidOntologyTerm(string id = null, string description = null, List<string> tags = null);
```
beforan, line 14: `IsValid()`

src/Directory/Services/Contracts/IOntologyTermService.cs
```diff
@@ -0,0 +1,24 @@
...
+        Task<bool> IsOntologyTermInUse(string id);
```
beforan, line 15: `IsInUse()`

src/Directory/Services/Contracts/IOntologyTermService.cs
```diff
@@ -0,0 +1,24 @@
...
+        
+        Task<int> CountOntologyTerms(string description = null, List<string> tags = null);
```
beforan, line 17: `Count()`

src/Directory/Services/Contracts/IOntologyTermService.cs
```diff
@@ -0,0 +1,24 @@
...
+        Task<int> GetCollectionCapabilityCount(string id);
```
beforan, line 18: This is a hard one, though partly cos without docs I don't 100% know what it does. I assume count Collecitons and Capabilities whose diagnosis is this term?

So something like `CountCollectionCapabilityByDiseaseStatus(string diseaseStatusId)` should make it a bit clearer? Not 100% sold on that though.

Def change the parameter name though, since the id doesn't directly relate to the types metnioned in the method name, so be explicit.

src/Directory/Services/Contracts/IOntologyTermService.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        Task DeleteOntologyTerm(string id);
```
beforan, line 20: `Delete()`

src/Directory/Services/Contracts/IOntologyTermService.cs
```diff
@@ -0,0 +1,24 @@
...
+        Task<OntologyTerm> UpdateOntologyTerm(OntologyTerm ontologyTerm);
```
beforan, line 21: `Update()`

src/Directory/Services/Contracts/IOntologyTermService.cs
```diff
@@ -0,0 +1,24 @@
...
+        Task<OntologyTerm> AddOntologyTerm(OntologyTerm ontologyTerm);
```
beforan, line 22: Not sure if I prefer `Create()` rather than `Add()` here? Kind of a consistency call as it sets the convention for future primarily entity typed services?

src/Directory/Services/OntologyTermService.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            _db = db;
+            _indexService = indexService;
+        }
+
+        protected IQueryable<OntologyTerm> QueryOntologyTerms(
+            string id = null, string description = null, List<string> tags = null, bool onlyDisplayable = false)
+        {
+            var query = _db.OntologyTerms
+                .AsNoTracking()
```
beforan, line 29: Be careful with this. I might change the method name to `ReadOnlyQuery()` or something, to be clear that this particular IQueryable has tracking off.

Or parameterise tracking and default to off...

Just don't want people accidentally using it for write activity and wondering why it's not tracking.

(This is one advantage of the old separate Read and Write services; the Read Service *never* needs tracking)

src/Directory/Services/OntologyTermService.cs
```diff
@@ -0,0 +1,136 @@
...
+            currentTerm.DisplayOnDirectory = ontologyTerm.DisplayOnDirectory;
+            currentTerm.MaterialTypes = ontologyTerm.MaterialTypes;
+            currentTerm.SnomedTag = ontologyTerm.SnomedTag;
+
+            await _db.SaveChangesAsync();
+
+            await _indexService.UpdateCollectionsOntologyOtherTerms(ontologyTerm.Value);
+            await _indexService.UpdateCapabilitiesOntologyOtherTerms(ontologyTerm.Value);
+
+            // TODO: Check If Manual Update Of Material Types Is Required
```
beforan, line 118: has this been resolved?

beforan (changes requested):

src/Directory/Services/Contracts/IOntologyTermService.cs
beforan: but the term in `CRUD` is `Create` so that's where I'm coming from

src/Directory/Services/Contracts/IOntologyTermService.cs
beforan: Also, I don't like `start` and `length`.

`offset` and `limit` are terms i've used in the past, and i think are common (though that may be anecdotally)

alternatively, since we're in c#, why not `skip` and `take` to match the underlying Linq.

src/Directory/Services/Contracts/IOntologyTermService.cs
Flemingjp: Gone with `Create` - don't know why but it sounds more permanent

beforan (dismissed):

src/Directory/Services/OntologyTermService.cs
Flemingjp: I have managed to resolve the issue for the Update method. However the Add method is still creating duplicate MaterialTypes rather than referencing existing MaterialTypes

Flemingjp:
As mentioned in #315 - there needs to be a distinct and clear difference in the service for where we want to retrieve OntologyTerms using the exact description or using it as a wildcard.

- Exact in almost all cases
- Wildcard for search cases (eg Search Type-Ahead)

beforan (approved):

blargly (approved):

PR #301, opened 2021-08-03
AB#38979: Organistation and Network Service Refactor

Flemingjp:
Refactored Organisation and Network methods into their own respective services, removing those methods from the `BiobankReadService` and `BiobankWriteService`.

The implementations use the `DbContext` directly, removing the use of the `GenericRespository` pattern.


src/Directory/Services/OrganisationService.cs
```diff
@@ -0,0 +1,583 @@
...
+            organisation = await Update(organisation.OrganisationId, org =>
+            {
+                org.OrganisationExternalId += org.OrganisationId;
+            });
+
+            return organisation;
+        }
+
+        /// <inheritdoc/>
+        public async Task<Organisation> Update(int id, Action<Organisation> updates)
```
Flemingjp, line 205: Thoughts on this pattern for Update? In refactoring I found the current way of mapping from `DTO`->`Untracked Entity`->`Tracked Entity` complex and problematic. 

This new pattern exposes a `Tracked Entity` in a limited scope such that updates can be made directly from a `DTO` . It also detaches the updated entity, such that it can return an untracked reference of the updated entity.

There we can rewrite a code block such as 
```c#
var biobankId = model.BiobankId;
var organisation = await _organisationService.Get(biobankId);

organisation.CollectionTypeId = model.CollectionType;
organisation.AccessConditionId = model.AccessCondition;

await _organisationService.Update(organisation);
```

Into
```c#
await _organisationService.Update(organisationId: model.BiobankId, organisation =>
{
  organisation.CollectionTypeId = model.CollectionType;
  organisation.AccessConditionId = model.AccessCondition;
});
```

It can also still support AutoMapper
```c#
await _organisationService.Update(organisationId: model.BiobankId, organisation =>
{
  _mapper.Map(model, organisation);
});
```

beforan: Upon discussion we've agreed that the current `Update()` approach was a flawed by-product of both the generic repository and the separate read/write services.

What we really need is to make Service methods more domain oriented, and have the service methods responsible for getting tracked entities and modifying them, rather than the pattern described above of getting an untracked entity, modifying it, then attaching it, then getting confused.

Rather than the above's approach of "fixing" that pattern, rather we should do away with it altogether and put more responsibility on the service methods, where they can decide whether or not they need a tracked entity.

src/Directory/Services/OrganisationService.cs
```diff
@@ -0,0 +1,507 @@
...
+
+        private async Task<Organisation> Suspend(int organisationId, bool suspend)
+        {
+            var organisation = await Update(new OrganisationDTO
+            {
+                OrganisationId = organisationId,
+                IsSuspended = suspend
+            });
+
+            //await _indexService.BulkIndexBiobank(organisation);
```
Flemingjp, line 321: I'm not sure of the logic here - does this indexing call need to be made? The Organisation is indexed via the prior `Update` method

Flemingjp: No need to index at this level, as all indexing is handled via the Update method

src/Directory/Biobanks.Web/Windsor/WindsorInstaller.cs
```diff
@@ -104,10 +106,6 @@ public void Install(IWindsorContainer container, IConfigurationStore store)
                     .ImplementedBy<ApplicationUserStore>()
                     .LifeStyle.Transient,
 
-                Component.For(typeof(IOntologyTermService))
```
beforan, line 107: is removing this in this PR correct?

Flemingjp: It is not

blargly (approved):

beforan (approved):

PR #305, opened 2021-08-12
AB#39541: Validate Organisation To Check For Default Aggregation Values

Flemingjp:
For a collection to be properly aggregated, the organsiation must set up the default aggregation values for `AccessCondition` and `CollectionType`. This is configured in the Directory UI under the Bulk Submissions settings.

- Users cannot generate API credentials until default values are configured
- Aggregator will skip aggregating a collection if the Organisation does not have configured values

### DIFF
--- a/src/Core/Jobs/AggregatorJob.cs
+++ b/src/Core/Jobs/AggregatorJob.cs
@@ -84,9 +84,19 @@ namespace Core.Jobs
                 throw new ArgumentException("Supplied baseSample is null, or is referencing a non-existent SampleContent");
 
             var organisation = await _organisationService.GetById(baseSample.OrganisationId);
+            var collectionName = _aggregationService.GenerateCollectionName(baseSample);
+
+            // Check Organisation Has Properly Configured Default Values
+            if (organisation.AccessCondition is null || organisation.CollectionType is null)
+            {
+                // Log Warning
+                Console.WriteLine($"Could not aggregate {collectionName} for {organisation.Name} due to missing default aggregation values.");
+
+                // Return As We Can Not Aggregate These Samples
+                return;
+            }
 
             // Find Exisiting Or Generate New Collection
-            var collectionName = _aggregationService.GenerateCollectionName(baseSample);
             var collection = await _collectionService.GetCollection(organisation.OrganisationId, collectionName);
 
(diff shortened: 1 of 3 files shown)


src/Directory/Biobanks.Web/Views/Biobank/Submissions.cshtml
```diff
@@ -2,162 +2,174 @@
...
-                class="btn btn-success pull-right">
-            Generate New Client Secret
-        </button>
-    </div>
-</div>
+@* API Credentials *@
+<h3>Submission API Authentication</h3>
 
+@* API Client Credential Only Accessible Once Organisation Defaults Are Set *@
+@if (Model.AccessCondition != null && Model.CollectionType != null)
```
Flemingjp, line 107: This section is now conditional

src/Core/Jobs/AggregatorJob.cs
```diff
@@ -84,9 +84,19 @@ private async Task AggregateCollectionSamples(LiveSample baseSample, IEnumerable
                 throw new ArgumentException("Supplied baseSample is null, or is referencing a non-existent SampleContent");
 
             var organisation = await _organisationService.GetById(baseSample.OrganisationId);
+            var collectionName = _aggregationService.GenerateCollectionName(baseSample);
+
+            // Check Organisation Has Properly Configured Default Values
+            if (organisation.AccessCondition is null || organisation.CollectionType is null)
+            {
+                // Log Warning
+                Console.WriteLine($"Could not aggregate {collectionName} for {organisation.Name} due to missing default aggregation values.");
```
Flemingjp, line 93: Not sure on the best way to expose a warning via Hangfire. Looking around resulted in an extension to Hangfire https://github.com/pieceofsummer/Hangfire.Console

beforan: we need to pursue the logging plugin, but that's kind of its own pbi.

This behaviour increases its priority though, i would say.

Flemingjp: OK I would suggest keeping it as this for time being, but with a future aim to implement Hangfire logging

JosephLavagna (approved):

blargly (approved):

PR #313, opened 2021-08-31
AB#38967: Refactor Publication Service

Flemingjp:
Migrated all Publication service methods to dedicated PublicationService


src/Directory/Services/PublicationService.cs
```diff
@@ -0,0 +1,59 @@
...
+        public async Task<Publication> Create(Publication publication)
+        {
+            _db.Publications.Add(publication);
+            await _db.SaveChangesAsync();
+
+            return publication;
+        }
+
+        ///<inheritdoc/>
+        public async Task<Publication> Update(string publicationId, int organisationId, Action<Publication> updates)
```
Flemingjp, line 39: See comment on #301

src/Directory/Services/PublicationService.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            var publication = await _db.Publications
+                .FirstOrDefaultAsync(x => x.PublicationId == publicationId && x.OrganisationId == organisationId);
+
+            if (publication is null)
```
Flemingjp, line 44: Do we just return null, or is an exception preferred?

src/Directory/Services/PublicationService.cs
```diff
@@ -0,0 +1,55 @@
...
+            return publication;
+        }
+
+        ///<inheritdoc/>
+        public async Task<Publication> Claim(string publicationId, int organisationId, bool accept = true)
+        {
+            var publication = await _db.Publications
+                .FirstOrDefaultAsync(x => x.PublicationId == publicationId && x.OrganisationId == organisationId);
+
+            if (publication is null)
```
Flemingjp, line 45: Do we want to return null or `KeyNotFoundException`

src/Directory/Services/PublicationService.cs
```diff
@@ -0,0 +1,55 @@
...
+                return null;
+
+            publication.Accepted = accept;
+
+            await _db.SaveChangesAsync();
+
+            return publication;
```
Flemingjp, line 52: Publication has no DTO - so this needs to be untracked ideally or we need to create a DTO

beforan: i don't think it's the end of the world that it's tracked now (though it's less consistent between layers)

if you wanted to reuse the entity as if it were a dto (and therefore untracked) you could always instantiate a new one and use automapper to apply the property values from the tracked one to the untracked return one.

src/Directory/Services/PublicationService.cs
beforan: i think KeyNotFound is better here.

It's kind of about semantics:
  - If you ask a service to find something (e.g. a `Get()` method), it's acceptable for it to return nothing if unsuccessful
  - If you ask a service to modify something, it's kind of disingenuous for it to return nothing when it didn't find it. Instead it should inform you it couldn't find the item in question and therefore couldn't proceed with what you asked it to do.

beforan (approved):

blargly (approved):

PR #314, opened 2021-09-06
AB#38959: Reference Data Service

Flemingjp:
Many reference data service methods currently achieve the same CRUD operations. Therefore refactoring into a generic reference data service will allow for a DRY approach to reference data services.

This PR creates a `ReferenceDataBase` entity and `ReferenceDataService<T>` implementation that can facilitate the refactoring of all the existing reference data service methods.

Changes include:
- All reference data entities now extend `ReferenceDataBase` standardising their basic properties to `Id`, `Value` and `SortOrder` 
- A migration to facilitate any changes to existing reference data schema
- `IReferenceDataService<T>` and `ReferenceDataService<T>` which can be extended per reference data where needed


src/Directory/Services/Contracts/IReferenceDataService.cs
```diff
@@ -0,0 +1,23 @@
...
+        Task Delete(int id);
+
+        Task<bool> Exists(string value);
+
+        Task<T> Get(int id);
+
+        Task<ICollection<T>> List();
+
+        Task<T> Update(T entity);
+    }
```
Flemingjp, line 22: There is scope for two additional methods `Task<int> GetUsageCount(int id)` and `Task<bool> IsInUse(int id)` 

The issue here is that these methods, while are applicable to all/most ReferenceData types, do not have a common generic implementation, as the usage typically depends on another entity. For example, for `AccessCondition` usage depends on the number of Collections it is used in.

I think the options are

- These methods are included with a default implementation
  - Cons: Requires sensible defaults that may not exist `(Usage = 0, InUse=false)` are those sensible? 
- These methods are included as abstract methods, making the base service abstract
  - Cons: Then requires an implementing class per each reference data type
- These methods are not included
  - Cons: Requires a specific implementation per each reference data type that requires these methods

src/Directory/Services/ReferenceDataService.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            await _db.SaveChangesAsync();
+        }
+
+        /// <inheritdoc/>
+        public async Task<bool> Exists(int id)
+            => await Query().AnyAsync(x => x.Id == id);
+
+        /// <inheritdoc/>
+        public async Task<bool> ExistsByValue(string value)
```
Flemingjp, line 59: `ExistsByValue` or just `Exists` as an method overload?

beforan: vastly prefer overload myself

src/Directory/Services/Contracts/IReferenceDataService.cs
beforan: default implementation no. it's too dangerous.

abstract vs exclude depends on 2 things, I think:
- How many concrete implementations do we save ourselves by not going the abstract route. My guess is not many, but that will be a factor as to whether or not it's "worth" not doing.
- How meaningful is it for stuff higher up to know that it can ask a service for usage information? Does it have to opt into that by a specific service type for a specific refdata type, just like bounds and other extended functionality? Or should it be reasonably assumed that you can ask that information about all refdata. Actually, this is probably largely informed by the point above ¯\_(ツ)_/¯

My gut feel, though subject to the above, is that abstract is the Right Thing To Do™, despite resulting in needing concrete implementations for all refdata.

At the end of the day, we are still using inheritance, so those multiple implementations are still more concise and DRY than they would otherwise have been.

beforan (changes requested):

beforan (approved):

blargly (approved):

PR #359, opened 2022-01-18
AB#48791: FIND Collection Associated Data Type Notes

JosephLavagna:
Added text box to the add collection page underneath the associated data types heading.
Added config value to turn on and off
Added seperate content page for details on Neglected Tropical Diseases (NTD)

### DIFF
--- a/src/Directory/Data/BiobanksDbInitializer.cs
+++ b/src/Directory/Data/BiobanksDbInitializer.cs
@@ -77,20 +77,6 @@ namespace Biobanks.Directory.Data
                 }
             };
 
-            var NTDConfig = context.Configs.Where(x => x.Key.Contains(ConfigKey.CollectionsNotes)).FirstOrDefault();
-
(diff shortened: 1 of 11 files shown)


src/Directory/Data/BiobanksDbInitializer.cs
```diff
@@ -77,6 +77,20 @@ private void SeedContentPages(BiobanksDbContext context)
                 }
             };
 
+            var NTDConfig = context.Configs.Where(x => x.Key.Contains(ConfigKey.CollectionsNotes)).FirstOrDefault();
+
+            if (NTDConfig.Value != "false")
+            {
+                pages.Add(new ContentPage
+                {
+                    Title = "Neglected Tropical Disease Associated Data Types",
```
blargly, line 86: I don't think we should be seeding content pages in the main directory code, as this is specific to Find only

blargly (approved):

PR #371, opened 2022-03-01
AB#51374: Modified the biobank activity report to have a side panel

ShashiprabhaRajapaksha:
Implementations of side panels in ASP.Net Razor views

### DIFF
--- a/src/Directory/Biobanks.Web/Content/custom.css
+++ b/src/Directory/Biobanks.Web/Content/custom.css
@@ -1,0 +1,4 @@
+.vl {
+    border-right: 1px solid grey;
+    background-color: lightgrey;
+}
--- a/src/Directory/Biobanks.Web/Views/ADAC/BiobankActivity.cshtml
+++ b/src/Directory/Biobanks.Web/Views/ADAC/BiobankActivity.cshtml
@@ -7,11 +7,34 @@
 
 @Html.Partial("_ADACTabs", "BiobankActivity")
 
-<h3>@ViewBag.Title</h3>
-
 <div class="row">
-    <div class="col-sm-12">
-        <br />
+
+    <style>
+        .vl {
+            border-right: 1px solid grey;
+            background-color: lightgrey;
+            }
+    </style>
+
+    <div class="col-sm-2">
+
+        <div class="sidebar-nav">
+
+            <ul class="nav flex-column vl">
+
+                <li class="nav-item">
+                    <a class="nav-link" href="#">Biobanks Activity</a>
+                </li>
+
+            </ul>
+
+        </div>
+
+    </div>
+    <div class="col-sm-9">
+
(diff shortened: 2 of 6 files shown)


blargly (approved):

aydtmiri (approved):

src/Directory/Biobanks.Web/Views/ADAC/BiobankActivity.cshtml
```diff
@@ -7,11 +7,34 @@
 
 @Html.Partial("_ADACTabs", "BiobankActivity")
 
-<h3>@ViewBag.Title</h3>
-
 <div class="row">
-    <div class="col-sm-12">
-        <br />
+
+    <style>
```
beforan, line 12: we don't want an inline `<style>` tag in the view - that's what we have CSS files for.

I see the same styles are defined in `custom.css`, so we should just be using that.

src/Directory/Biobanks.Web/Content/custom.css
```diff
@@ -0,0 +1,4 @@
+.vl {
```
beforan, line 1: The CSS here is fine, but there's just a few improvements:

- Could we have a comment to give the intended usage? I guess `vl` is for "Vertical Line" or similar. Would be good to just have a not saying that, and that it's intended to be used a vertical divider on divs.
- `custom.css` file isn't very clearly named. For general stuff it'd be better to put it in the existing `Site.css` file, or for specific stuff you could use a new CSS file but give it a specific name (e.g. the `nav.css` file is exclusively for the nav area)
- Please move the new CSS file to inside the `Site/` folder in `Content/` as that's where, in this project, we keep CSS we wrote ourselves, as opposed to stuff dumped there by libraries (e.g. bootstrap etc.)

beforan (changes requested):

src/Directory/Biobanks.Web/Content/custom.css
beforan: Also, currently this CSS file isn't being used.

In ASP.NET MVC (before ASP.NET Core) there is a principle of "bundles" for CSS and JS files. We define a `StylesBundle` [here](https://github.com/biobankinguk/biobankinguk/blob/3dfcda64eb2f12a85e6f7eebd1c231a656eccd08/src/Directory/Biobanks.Web/App_Start/BundleConfig.cs#L17) which you can see includes all CSS files inside `Site/`. So if you move you5r new CSS file, it will get used.

src/Directory/Biobanks.Web/Views/ADAC/BiobankActivity.cshtml
beforan: I guess this might have been added because the CSS file wasn't loading?

If so, fixing the CSS file loading as per my other comment will mean this can be removed.

PR #376, opened 2022-03-04
AB#46818: Omop context

aydtmiri:
Added Omop Context


src/Omop/Omop/Context/OmopDbContext.cs
```diff
@@ -0,0 +1,56 @@
...
+            model.Entity<ConditionOccurence>()
+                .HasIndex(x => new
+                {
+                    x.ConditionOccurenceId
+
+                }).IsUnique();
+        }
+
+
+        //public class OmopDbContextFactory : IDesignTimeDbContextFactory<OmopDbContext>
```
beforan, line 32: All this commented code can go. We don't need it since we aren't running migrations against this context

src/Omop/Omop/Entities/ConditionOccurence.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+
+namespace Biobanks.Omop.Entities
+{
+    /// <inheritdoc />
+    public class ConditionOccurence
+    {
+        public int ConditionOccurenceId { get; set; }
```
beforan, line 8: is this actually a primary key? If so, can we use the `[Key]` attribute on it, and then you don't need the unique constraint configured on the context either?

If not, then fair enough :P

beforan (changes requested):

blargly (approved):

AndyRae (approved):

PR #386, opened 2022-05-23
AB#47303: Added related associated data in 'add collection'

Samad198:
## Overview

Changed behaviour of associated data list inside 'add collection' page to add and remove data depending on Disease status input

## Azure Boards

- [AB#47303](https://dev.azure.com/UniversityOfNottingham/bd2de5ba-2a41-4bdc-a444-61e6b706132a/_workitems/edit/47303)
- [AB#47304](https://dev.azure.com/UniversityOfNottingham/bd2de5ba-2a41-4bdc-a444-61e6b706132a/_workitems/edit/47304)

### DIFF
--- a/src/Directory/Biobanks.Web/Controllers/BiobankController.cs
+++ b/src/Directory/Biobanks.Web/Controllers/BiobankController.cs
@@ -958,6 +958,52 @@ namespace Biobanks.Web.Controllers
             return View((AddCollectionModel)(await PopulateAbstractCRUDCollectionModel(new AddCollectionModel { FromApi = false })));
         }
 
+        [HttpGet]
+        public async Task<ActionResult> GetAssociatedDataTypeViews(string id)
+        {
+            AddCollectionModel model = new AddCollectionModel { FromApi = false };
+
+            var timeFrames = (await _associatedDataProcurementTimeframeService.List())
+               .Select(x => new AssociatedDataTimeFrameModel
+               {
+                   ProvisionTimeId = x.Id,
+                   ProvisionTimeDescription = x.Value,
+                   ProvisionTimeValue = x.DisplayValue
+               });
+
+            var types = (await _ontologyTermService.ListAssociatedDataTypesByOntologyTerm(id))
+                     .Select(x => new AssociatedDataModel
+                     {
+                         DataTypeId = x.Id,
+                         DataTypeDescription = x.Value,
+                         DataGroupId = x.AssociatedDataTypeGroupId,
+                         Message = x.Message,
+                         TimeFrames = timeFrames,
+                         isLinked = true
+        });
+            model.Groups = new List<AssociatedDataGroupModel>();
+            var groups = await _associatedDataTypeGroupService.List();
+
+            foreach (var g in groups)
+            {
+                var groupModel = new AssociatedDataGroupModel();
+                groupModel.GroupId = g.Id;
+                groupModel.Name = g.Value;
+                groupModel.Types = types.Where(y => y.DataGroupId == g.Id).ToList();
+                
+                model.Groups.Add(groupModel);
+            }
+
+            //Check if types are valid
+            foreach (var type in types)
+            {
+                type.Active = model.AssociatedDataModelsValid();
+            }
+
+            return PartialView("_LinkedAssociatedData", model);
+
+        }
+
         [HttpPost]
         [ValidateAntiForgeryToken]
         public async Task<ActionResult> AddCollection(AddCollectionModel model)
--- a/src/Directory/Biobanks.Web/Controllers/BiobankController.cs
+++ b/src/Directory/Biobanks.Web/Controllers/BiobankController.cs
@@ -1524,7 +1524,8 @@ namespace Biobanks.Web.Controllers
         }
 
         private async Task<AbstractCRUDCapabilityModel> PopulateAbstractCRUDAssociatedData(
-            AbstractCRUDCapabilityModel model)
+            AbstractCRUDCapabilityModel model,
+            string ontologyTermId = null)
         {
             var timeFrames = (await _associatedDataProcurementTimeframeService.List())
                 .Select(x => new AssociatedDataTimeFrameModel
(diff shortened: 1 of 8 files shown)


src/Directory/Biobanks.Web/Controllers/BiobankController.cs
```diff
@@ -1524,7 +1524,8 @@ public async Task<ViewResult> SampleSet(int id)
         }
 
         private async Task<AbstractCRUDCapabilityModel> PopulateAbstractCRUDAssociatedData(
-            AbstractCRUDCapabilityModel model)
+            AbstractCRUDCapabilityModel model,
+            string ontologyTermId = null)
```
beforan, line 1528: You don't seem to have updated anywhere this is _called_ so `ontologyTermId` will always be `null` and your new code will never run?

Also, I'm not sure a new argument passing ontologyTermId is the right way to do this? It's certainly not clear what the effect is going rto be, and the `AbstractCRUDCapabilityModel` already has the property `Diagnosis` containing a Diagnosis restricted OntologyTerm value, so you can get the ID the same way other methods do from a model taht only contains a value: `var ontologyTerm = await _ontologyTermService.Get(value: model.Diagnosis);` (e.g. [in AddCollection](https://github.com/biobankinguk/biobankinguk/blob/a8c166dee135021a8fee37eb7719acf9ef516f83/src/Directory/Biobanks.Web/Controllers/BiobankController.cs#L989)

src/Directory/Biobanks.Web/Scripts/Shared/diagnosis-type-ahead.js
```diff
@@ -1,45 +1,135 @@
...
+                                        ${assData.Value}
+                                    </label>
+                                    <input data-val="true" data-val-number="The field DataTypeId must be a number." data-val-required="The DataTypeId field is required." id="Groups_0__Types_0__DataTypeId" name="Groups[0].Types[0].DataTypeId" type="hidden" value="3">
+                                </div>
+                            </td>
+                            <td>
+                                <div class="radio">
+                                        <label class="timeFrames">
+                                            <input class="ass-dat-rad" data-val="true" data-val-number="The field ProvisionTimeId must be a number." id="Groups_0__Types_0__ProvisionTimeId" name="Groups[0].Types[0].ProvisionTimeId" type="radio" value="1">
+                                            Immediate
```
beforan, line 85: These timeframes can't be hardcoded. They're based on database records, so you need to render on the page what comes out of the db for timeframes.

The server side views do this, so the dynamic client side bits need to too, I'm afraid.

On the server side MVC - the ViewModel and [EditorTemplate](https://github.com/biobankinguk/biobankinguk/blob/main/src/Directory/Biobanks.Web/Views/Biobank/EditorTemplates/AssociatedDataModel.cshtml) particularly - does lots of the heavy lifting in terms of generating the markup.

Perhaps this JS should ask the server for markup, and then add that to the page? We do this elsewhere, e.g. [here we render a Partial View in an "ajax endpoint" and it gets inserted into a modal i believe](https://github.com/biobankinguk/biobankinguk/blob/a8c166dee135021a8fee37eb7719acf9ef516f83/src/Directory/Biobanks.Web/Controllers/BiobankController.cs#L665)

src/Directory/Biobanks.Web/Views/Biobank/_CollectionSharedFields.cshtml
```diff
@@ -9,7 +9,7 @@
         <span id="support-email" data-email="@ConfigurationManager.AppSettings["AdacSupportEmail"]"></span>
     </label>
     <div class="col-sm-8">
-        @Html.TextBoxFor(x => x.Diagnosis, new { @class = "form-control diagnosis-search", autocomplete = "off" })
+        @Html.TextBoxFor(x => x.Diagnosis, new { @class = "form-control diagnosis-search",@id="diagnosis-textbox", autocomplete = "off" })
```
beforan, line 12: any reason to add this `id`? can't you just select on the `diagnosis-search` class? that seems to be what other code was already doing

beforan (changes requested):

src/Directory/Biobanks.Web/Views/Biobank/_CollectionSharedFields.cshtml
Samad198: Just removed it. Was no need for it

src/Directory/Biobanks.Web/Scripts/Shared/diagnosis-type-ahead.js
Samad198: Have now added this

src/Directory/Biobanks.Web/Controllers/BiobankController.cs
Samad198: Seems i was confused on what I needed to do. Have removed it now

blargly (approved):

src/Directory/Biobanks.Web/Controllers/BiobankController.cs
```diff
@@ -958,6 +958,52 @@ public async Task<ViewResult> AddCollection()
             return View((AddCollectionModel)(await PopulateAbstractCRUDCollectionModel(new AddCollectionModel { FromApi = false })));
         }
 
+        [HttpGet]
+        public async Task<ActionResult> GetAssociatedDataTypeViews(string id)
```
beforan, line 962: Can you put `Ajax` on the end of this.

I know it's kinda dumb but it matches the convention used elsewhere in this app when there are Controller actions on non API Controllers intended for use in AJAX calls.

beforan (changes requested):

src/Directory/Biobanks.Web/Controllers/BiobankController.cs
Samad198: Done

beforan (approved):

blargly (approved):

PR #397, opened 2022-06-10
AB#47374: AssociatedDataOntologyTerms bug fixes

Samad198:
## Overview

Fixed bugs to do with associated data and ontology terms relationship

## Azure Boards

- AB#47374

## Notes

Everything to my knowledge seems to work except updating an existing associated data or ontology term to have more or less many to many relationships as it throws this error. 
image
Full message:
ExceptionMessage: "The operation failed: The relationship could not be changed because one or more of the foreign-key properties is non-nullable. When a change is made to a relationship, the related foreign-key property is set to a null value. If the foreign-key does not support null values, a new relationship must be defined, the foreign-key property must be assigned another non-null value, or the unrelated object must be deleted."

Will look further into how to fix on monday

### DIFF
--- a/src/Directory/Biobanks.Web/ApiControllers/DiseaseStatusController.cs
+++ b/src/Directory/Biobanks.Web/ApiControllers/DiseaseStatusController.cs
@@ -12,6 +12,8 @@ using System.Collections.Generic;
 using Biobanks.Directory.Services.Constants;
 using Biobanks.Directory.Services.Contracts;
 using Biobanks.Entities.Data.ReferenceData;
+using System.ComponentModel.DataAnnotations;
+using Newtonsoft.Json;
 
 namespace Biobanks.Web.ApiControllers
 {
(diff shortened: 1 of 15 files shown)


src/Directory/Biobanks.Web/ApiControllers/DiseaseStatusController.cs
```diff
@@ -139,8 +141,10 @@ public async Task<IHttpActionResult> Put(string id, OntologyTermModel model)
 
         [HttpPost]
         [Route("")]
-        public async Task<IHttpActionResult> Post(OntologyTermModel model)
+        public async Task<IHttpActionResult> Post([System.Web.Mvc.Bind(Exclude = "Value",Include ="OntologyTermId, Description",Prefix = "")] OntologyTermModel model)
```
blargly, line 144: Did we not decide that this bind command didn't work? I think it was just one of the attempted fixes?

blargly (changes requested):

src/Directory/Biobanks.Web/ApiControllers/DiseaseStatusController.cs
Samad198: Oh yeah I just forgot to remove it

Samad198: resolved

beforan (dismissed):

blargly (approved):

blargly (approved):

PR #430, opened 2022-10-04
AB#78716: AgeRange Controller

MoizIqbal97:
## Overview
Porting over AgeRange Controller
SwaggerUI Annotation

## Azure Boards

- [AB#78716](https://dev.azure.com/UniversityOfNottingham/bd2de5ba-2a41-4bdc-a444-61e6b706132a/_workitems/edit/78716)

### DIFF
--- a/src/Submissions/Api/Controllers/ReferenceData/AgeRangeController.cs
+++ b/src/Submissions/Api/Controllers/ReferenceData/AgeRangeController.cs
@@ -142,6 +142,11 @@ namespace Biobanks.Submissions.Api.Controllers.ReferenceData
                 ModelState.AddModelError("AgeRange", $"The age range \"{model.Value}\" is currently in use, and cannot be deleted.");
             }
 
+            if (!ModelState.IsValid)
+            {
+                return BadRequest(ModelState);
+            }
+
             await _ageRangeService.Delete(id);
 
             return Ok(id);
--- a/src/Submissions/Api/Controllers/ReferenceData/AgeRangeController.cs
+++ b/src/Submissions/Api/Controllers/ReferenceData/AgeRangeController.cs
@@ -1,12 +1,13 @@
 ﻿using Biobanks.Entities.Data.ReferenceData;
-using Biobanks.Shared.Services.Contracts;
 using Biobanks.Submissions.Api.Models.Shared;
+using Biobanks.Submissions.Api.Services.Directory.Contracts;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Swashbuckle.AspNetCore.Annotations;
 using System.Collections;
 using System.Linq;
 using System.Threading.Tasks;
+using System.Xml;
 
 namespace Biobanks.Submissions.Api.Controllers.ReferenceData
 {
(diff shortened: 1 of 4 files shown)


src/Submissions/Api/Startup.cs
```diff
@@ -239,7 +239,8 @@ public void ConfigureServices(IServiceCollection services)
                 .AddTransient<IGoogleAnalyticsReportingService, GoogleAnalyticsReportingService>()
 
                 .AddTransient<ISubmissionExpiryService, SubmissionExpiryService>()
-                .AddTransient<IRegistrationDomainService, RegistrationDomainService>();
+                .AddTransient<IRegistrationDomainService, RegistrationDomainService>()
+              .AddTransient<Services.Directory.Contracts.IReferenceDataService<AgeRange>, AgeRangeService>();
```
AndyRae, line 243: Do we want to move these service registrations to outside the directory config now?

src/Submissions/Api/Controllers/ReferenceData/AgeRangeController.cs
```diff
@@ -43,8 +44,260 @@ public async Task<IList> Get()
...
+                SortOrder = convertedModel.SortOrder,
+                LowerBound = convertedModel.LowerBound,
+                UpperBound = convertedModel.UpperBound
+            };
+
+
+            await _ageRangeService.Add(range);
+            await _ageRangeService.Update(range); // Ensure sortOrder is correct
+
+            return Ok(model);
```
AndyRae, line 127: Might be more useful to return a 202 `Accepted` - as the insert is async. (for all of these)

src/Submissions/Api/Controllers/ReferenceData/AgeRangeController.cs
```diff
@@ -43,8 +44,260 @@ public async Task<IList> Get()
                     .Result
                 )
                 .ToList();
-
             return models;
         }
+
+        [HttpPost]
+        [SwaggerResponse(200, Type = typeof(AgeRangeModel))]
```
AndyRae, line 51: This endpoint also return a 400 (line 88), so should add the SwaggerResponse annotation for that here as well. (For all of them)

src/Submissions/Api/Controllers/ReferenceData/AgeRangeController.cs
```diff
@@ -43,8 +44,260 @@ public async Task<IList> Get()
...
+            return Ok(model);
+        }
+
+        [HttpDelete("{id}")]
+        [SwaggerResponse(200, Type = typeof(AgeRangeModel))]
+        public async Task<IActionResult> Delete(int id)
+        {
+            var model = await _ageRangeService.Get(id);
+
+            // If in use, prevent update
```
AndyRae, line 136: This doesn't seem to prevent anything - the modelstate needs to be checked and return the bad request. See the other endpoints for the pattern.

src/Submissions/Api/Startup.cs
MoizIqbal97: Yes, I believe this way we can actually run and test the end points using Postman.
Without doing this step I get a `System.InvalidOperationException: Unable to resolve service for type` error

src/Submissions/Api/Controllers/ReferenceData/AgeRangeController.cs
MoizIqbal97: Ah yes thats a good point, will do 👍

src/Submissions/Api/Controllers/ReferenceData/AgeRangeController.cs
beforan: careful that changing this may affect any caller (e.g. in javascript)

beforan: also, it awaits the update, so actually no it's not returning before the write completes. it's async from a threading perspective (it yeilds the thread) but from the methods perspective its synchronous

beforan (approved):

AndyRae (approved):

beforan (approved):

PR #436, opened 2022-10-08
AB#33843: Net6

blargly:
Upgraded project to Net 6 as a precursor to implementing identy on the new directory.

Replaced startup/program launch with modern startup free approach.

Keen to hear opinions on how program.cs should be organised.


src/Submissions/Api/Api.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
AndyRae, line 4: Moving to 6 - might be an opportunity to use [ImplicitUsings](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#implicitusings);
Would be interested to hear opinions on effectively hiding many of the `using System;` directives, but would note in new 6 apps that this is enabled by default.

src/Submissions/Api/Api.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
```
beforan, line 4: need to make sure all other projects which target net5 are changed to net6 as well.

e.g. `Data`

src/Submissions/Api/Program.cs
```diff
@@ -1,39 +1,370 @@
...
+options.UseNpgsql("Omop"));
+
+builder.Services.AddDbContext<BiobanksDbContext>(options =>
+    options.UseSqlServer(connectionString,
+    sqlServerOptions => sqlServerOptions.CommandTimeout(300000000)));
+builder.Services.AddDatabaseDeveloperPageExceptionFilter();
+
+//identity
+builder.Services.AddDefaultIdentity<IdentityUser>(options => options.SignIn.RequireConfirmedAccount = true)
+    .AddEntityFrameworkStores<BiobanksDbContext>();
```
beforan, line 67: Does `BiobanksDbContext` also need updating to inherit from `IdentityDbContext` for this to work?

(That is a desirable change at some point anyway)

src/Submissions/Api/Program.cs
```diff
@@ -1,39 +1,370 @@
...
+builder.Services.AddDbContext<OmopDbContext>(options =>
+options.UseNpgsql("Omop"));
+
+builder.Services.AddDbContext<BiobanksDbContext>(options =>
+    options.UseSqlServer(connectionString,
+    sqlServerOptions => sqlServerOptions.CommandTimeout(300000000)));
+builder.Services.AddDatabaseDeveloperPageExceptionFilter();
+
+//identity
+builder.Services.AddDefaultIdentity<IdentityUser>(options => options.SignIn.RequireConfirmedAccount = true)
```
beforan, line 66: `IdentityUser` should be `ApplicationUser` or whatever the custom user class is called. I know it's been copied over to Submissions already.

src/Submissions/Api/Program.cs
```diff
@@ -1,39 +1,370 @@
...
+    .AddEntityFrameworkStores<BiobanksDbContext>();
+
+builder.Services.AddRazorPages();
```
beforan, line 69: Don't think Submissions uses Razor Pages, so don't add them

src/Submissions/Api/Program.cs
```diff
@@ -1,39 +1,370 @@
...
+
+                // Hangfire Server
+                .UseHangfireDashboard();
+
+app.UseRouting();
+
+app.UseAuthentication();
+app.UseAuthorization();
+
+app.MapRazorPages();
```
beforan, line 368: As above, don't believe this is needed.

src/Submissions/Api/appsettings.json
```diff
@@ -18,11 +18,7 @@
     "EntitiesPerSubmission": 10000
   },
 
-  "Logging": {
-    "LogLevel": {
-      "Default": "Warning"
-    }
-  },
+"Microsoft.AspNetCore":  "Warning",
```
beforan, line 21: This seems wrong... pretty sure it should still be inside a logging section?

beforan (changes requested):

src/Submissions/Api/Api.csproj
beforan: I would suggest as a starter for 10 we do what the net6 templates do by default for this.

But yeah, agree could be useful going forwards

src/Submissions/Api/Program.cs
beforan: I appreciate that the actual identity config is still to come in a later PR, but using the wrong class here at this stage might cause undesirable extra migrations to get the right Users table schema.

AndyRae (dismissed):

AndyRae (approved):

MoizIqbal97 (approved):

PR #453, opened 2022-10-14
AB#79908: Fix build warnings

AndyRae:
## Overview

Fixes misc build warnings across the new submissions api. These are mostly xml warnings and qol improvements.

The main change is moving the application insights telemtry to a singleton pattern, as the framework approach has been [deprecated](https://github.com/microsoft/ApplicationInsights-dotnet/issues/1152), and would cause memory leaks.

This also means using telemetry is consistent with other services, and can be injected when necesssary inside services.

### DIFF
--- a/src/Entities/Data/Collection.cs
+++ b/src/Entities/Data/Collection.cs
@@ -28,6 +28,7 @@ namespace Biobanks.Entities.Data
 
         public bool FromApi { get; set; }
 
+        // TODO: Remove the nullable context when the whole project supports .net 5+.
         #nullable enable
         public string? Notes { get; set; }
         #nullable disable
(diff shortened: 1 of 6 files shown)


src/Entities/Data/Collection.cs
```diff
@@ -28,7 +28,9 @@ public class Collection
 
         public bool FromApi { get; set; }
 
+        #nullable enable
         public string? Notes { get; set; }
```
beforan, line 32: Can we add a TODO to revisit non-nullable reference types once all the code is on a language version that supports them

This fix is fine for now given that both Framework and net5/6 code uses this project, but it's not the ideal solution once the framework code is gone

beforan (changes requested):

beforan (approved):

PR #461, opened 2022-10-26
AB#79744: Disease status controller

MoizIqbal97:
**Ports over the Disease Status controller
Creates a new service and interface**

The previous controller had dependencies on the BioBankReadService, which in turn had dependencies on identity, and does not work in core.
This PR removes this dependency, by implementing a new interface `IDiseaseStatusService.cs`


src/Submissions/Api/Services/Directory/Contracts/IDiseaseStatusService.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Biobanks.Entities.Shared.ReferenceData;
+using System.Threading.Tasks;
+
+namespace Biobanks.Submissions.Api.Services.Directory.Contracts
+{
+    public interface IDiseaseStatusService
+    {
+        Task<SnomedTag> GetSnomedTagByDescription(string description);
```
AndyRae, line 8: Just to flag for the future - this method isn't unique to the Disease Status service, so extracting it from the read service means we'll be duplicating it in other services that use it as well.

AndyRae (approved):

src/Submissions/Api/Services/Directory/Contracts/IDiseaseStatusService.cs
beforan: Is there sense in it going in a `SnomedTagService` now, that all of those cases can then use?

AndyRae: Yes to be specific - there's just one other service this is used by, the `ExtractionProcedureController`. So it's not as if we're handling lots more, so it may be okay to duplicate just once?

I'd say a new Snomed service is needed if we're expecting to add more of these snomed related services in the future (I don't understand the system enough to answer).

beforan: I can't remember what other SnomedTag functionality there might be, but if there's any, it'll make sense to have a service just for it.

Or lump it in with any other OntologyTerm stuff in an OntologyTermService. SnomedTags are a weird specific extension we use in relation to OntologyTerms; where DiseaseStatus is just one use of OntologyTerms, so i don't think the controller specific service is the best way to go anyway.

AndyRae: Yep I don't think we'd regret a new `SnomedTagService` then, and extract all the snomed related methods from the read service entirely.

AndyRae: Moved out `GetSnomedTagByDescription` and other snomed methods into `OntologyTermService` in #472.

beforan: We'll let #472 pick up that change then, and merge this one as is 👍

beforan (approved):

PR #463, opened 2022-11-01
AB#76206: Home controller

MoizIqbal97:
Adds ConfigKey string varliables
Contact Controller moved out to its own Controller
Home Controller added
Config Service updated to use IMemoryCache
_FeedbackMessage view ported


src/Submissions/Api/Controllers/HomeController.cs
```diff
@@ -0,0 +1,47 @@
...
+                NetworkRegistration = await _configService.GetSiteConfigValue(ConfigKey.HomepageNetworkRegistration, true, ""),
+                NetworkRegistration2 = await _configService.GetSiteConfigValue(ConfigKey.HomepageNetworkRegistration2, true, ""),
+                RequireSamplesCollected = await _configService.GetSiteConfigValue(ConfigKey.HomepageSearchRadioSamplesCollected, true, ""),
+                AccessExistingSamples = await _configService.GetSiteConfigValue(ConfigKey.HomepageSearchRadioAccessSamples, true, ""),
+                FinalParagraph = await _configService.GetSiteConfigValue(ConfigKey.HomepageFinalParagraph ,true,""),
+                ResourceRegistrationButton = await _configService.GetSiteConfigValue(ConfigKey.RegisterBiobankTitle, true, ""),
+                NetworkRegistrationButton = await _configService.GetSiteConfigValue(ConfigKey.RegisterNetworkTitle, true, "")
+            });
+        }
+      //  public ActionResult Cookies() => View(); TODO copy views over
```
AndyRae, line 45: If we're not adding the home / contact views at this point, it's worth adding a PBI to make sure we do pick them up.

AndyRae (approved):

src/Submissions/Api/Controllers/ContactController.cs
```diff
@@ -0,0 +1,157 @@
...
+
+namespace Biobanks.Submissions.Api.Controllers
+{
+    public class ContactController : Controller
+    {
+        private readonly INetworkService _networkService;
+        private readonly IOrganisationDirectoryService _organisationService;
+        private readonly IMapper _mapper;
+        private readonly IMemoryCache _memoryCache;
+        //private readonly IEmailService _emailService; //TODO: Email Service has not be ported yet
```
dcl10, line 24: ⛏️ If you're not using the commented code in this file anymore, just delete it

src/Submissions/Api/Controllers/HomeController.cs
```diff
@@ -0,0 +1,47 @@
...
+    {
+        private readonly IConfigService _configService;
+
+        public HomeController(IConfigService configService)
+        {
+            _configService = configService;
+        }
+
+        // GET: Home
+        public async Task<IActionResult> IndexAsync()
```
dcl10, line 24: Does this not need a `[HttpGet]` decorator?

src/Submissions/Api/Controllers/HomeController.cs
dcl10, line 45: ⛏️ remove if not needed

dcl10 (changes requested):

src/Submissions/Api/Controllers/HomeController.cs
MoizIqbal97: We are inheriting from the `Controller` class, the convention for this is `/<ControllerName>/<ActionMethodName>` .
Index method is a special action name where it is served from just `/<ControllerName>/` 

The decorates are only used for API controllers which uses the `ControllerBase` class and the `[ApiController]` attribute

(answer plagarised from @beforan XD )

dcl10 (approved):

AndyRae (approved):

PR #467, opened 2022-11-04
AB#83121: Emailservice

MoizIqbal97:
## Overview
EmailService and its views ported

## Notes

- HomeController moved to DirectoryFolder
- LocalDiskEmailSender and SendGridEmailSender updated to allow the use of `ccAddress`
- Minor refactoring of code


src/Submissions/Api/Controllers/ContactController.cs
```diff
@@ -1,8 +1,10 @@
 ﻿using AutoMapper;
```
AndyRae, line 1: Move this controller to `Api/Controllers/Directory` as well

MoizIqbal97: this is a general api controller rather than a directory controller

AndyRae (dismissed):

AndyRae (approved):

blargly (approved):

PR #480, opened 2022-11-14
AB#76209: Pages Admin Controller

MoizIqbal97:
Port over Page Controller and subsequent Views, Model and Script


blargly (approved):

src/Submissions/Api/Api.csproj
```diff
@@ -25,7 +25,9 @@
     <PackageReference Include="ClacksMiddlware" Version="2.1.0" />
     <PackageReference Include="Hangfire" Version="1.7.31" />
     <PackageReference Include="MarkdownDeep.NET.Core" Version="1.5.0.4" />
+    <PackageReference Include="Microsoft.AspNet.Web.Optimization" Version="1.1.3" />
     <PackageReference Include="MimeKit" Version="3.4.2" />
+    <PackageReference Include="MvcSiteMapProvider.MVC5.Core" Version="4.6.26" />
```
beforan, line 30: Don't add this at this time, we need to figure out a Core compatible solution to it separately.

src/Submissions/Api/Api.csproj
```diff
@@ -25,7 +25,9 @@
     <PackageReference Include="ClacksMiddlware" Version="2.1.0" />
     <PackageReference Include="Hangfire" Version="1.7.31" />
     <PackageReference Include="MarkdownDeep.NET.Core" Version="1.5.0.4" />
+    <PackageReference Include="Microsoft.AspNet.Web.Optimization" Version="1.1.3" />
```
beforan, line 28: You can't add this, it's an old framework asp.net package.

(see the `Microsoft.AspNet` prefix; not `Microsoft.AspNetCore`)

remove it and then we'll have to fix whatever code was wanting it

src/Submissions/Api/Api.csproj
```diff
@@ -67,6 +69,7 @@
     <Folder Include="Content\images\" />
     <Folder Include="Pages\" />
     <Folder Include="Repositories\" />
+    <Folder Include="Scripts\Bundles\" />
```
beforan, line 72: not sure what this is included for?

src/Submissions/Api/App_Start/BundleConfig.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System.Web.Optimization;
+
+namespace Biobanks.Submissions.Api.App_Start
+{
+    public static class BundleConfig
+    {
```
beforan, line 6: You can't use `BundleConfig.cs` in Core.

Remove this file.

If you want JS to work you need to look at the Framework directory's `BundleConfig.cs` to find out what files were included in the bundle you want, then add those files MANUALLY to the page.

e.g.

the script bundle in this file is called `~/bundles/pageadmin` but references multiple actual js files, so you'd have to add the following to the page:

```html
<script src="~/Scripts/Directory/directory-pages-admin.js"></script>
<script src="~/Scripts/bootbox.js"></script> <!-- for however many bootbox scripts there are -->
<script src="~/Scripts/Directory/directory-refdata-utility.js"></script>
```

beforan: Alternatively wait for me to sort out the bundles in a Core way.

src/Submissions/Api/Config/ConfigKey.cs
```diff
@@ -4,12 +4,15 @@ public static class ConfigKey
...
+        public const string DisplayAnalytics = "site.display.analytics";
+        public const string ShowPreservationPercentage = "site.display.preservation.percent";
+        public const string ShowCounties = "site.display.counties";
 
         // Ref Data Name Configuration Options
         public const string MacroscopicAssessmentName = "site.display.macroscopicassessment.name";
         public const string DonorCountName = "site.display.donorcount.name";
         public const string StorageTemperatureName = "site.display.storagetemperature.name";
-        public const string ShowPreservationPercentage = "true";
+        public const string ShowPreservationPercentageBool = "true";
```
beforan, line 15: this file is a list of keys. "true" is not going to be the key for this config settings. it might be the value.

I suggest looking at the framework directory code to see what the key should be.

beforan: presumably something similar to the others, e.g. `site.display.showpreservationpercentagebool`

though I really hope it doesn't have `bool` on the end, that's not good naming practice.

src/Submissions/Api/Controllers/Submissions/PagesAdminController.cs
```diff
@@ -0,0 +1,167 @@
+﻿using Biobanks.Submissions.Api.Models.Submissions;
+using Biobanks.Submissions.Api.Services.Directory.Contracts;
+using Biobanks.Submissions.Api.Utilities;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using MvcSiteMapProvider.Web.Mvc.Filters;
```
beforan, line 6: I wouldn't expect a controller to reference MVC Sitemap, but if you're unsure how to take out what's using it, give me a shout. We can't use it as is in Core

beforan (changes requested):

src/Submissions/Api/Api.csproj
beforan: AB#84952 raised to track

beforan (approved):

AndyRae (approved):

PR #488, opened 2022-12-02
AB#84363: Biobanks Controller

MoizIqbal97:
Ports over the Biobanks Controller and Views
Configure Admin areas


src/Submissions/Api/Models/Biobank/BiobankAdminsModel.cs
```diff
@@ -0,0 +1,14 @@
+using Biobanks.Submissions.Api.Models.Shared;
```
AndyRae, line 1: This could go in `Area/Admin/Models` maybe?  
As far as I can tell this isn't a shared model - but solely used by the admin area Biobank controller, so would make sense to group it all in the area where possible.

Not crucial though

AndyRae (approved):

src/Submissions/Api/Models/Biobank/BiobankAdminsModel.cs
MoizIqbal97: Done :)

AndyRae (dismissed):

src/Submissions/Api/Config/SitePropertiesOptions.cs
```diff
@@ -16,6 +16,10 @@ public record SitePropertiesOptions
 
     public string LegalEntity { get; init; } = "Legal Entity";
 
+    public string EmailFromName { get; init; } = "EmailFromName";
```
beforan, line 19: this should be coming from email sender options, we don't need to add it to siteproperties as well.

src/Submissions/Api/Config/SitePropertiesOptions.cs
```diff
@@ -16,6 +16,10 @@ public record SitePropertiesOptions
...
+
+    public string EmailFromAddress { get; init; } = "EmailFromAddress";
```
beforan, line 21: same

src/Submissions/Api/Area/Admin/Controllers/BiobankController.cs
```diff
@@ -0,0 +1,210 @@
...
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Http.Extensions;
+using Microsoft.AspNetCore.Identity;
+using Microsoft.AspNetCore.Mvc;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Biobanks.Submissions.Api.Area.Admin.Controllers;
+public class BiobankController : Controller
```
beforan, line 20: - Controller names should be plural - change the controller and file name to `BiobanksController`
- Controllers inside Areas need annotating with the Area as per [this](https://learn.microsoft.com/en-us/aspnet/core/mvc/controllers/areas?view=aspnetcore-7.0#associate-the-controller-with-an-area)
- Can we use the default conventional path for Areas as per [this](https://learn.microsoft.com/en-us/aspnet/core/mvc/controllers/areas?view=aspnetcore-7.0#area-folder-structure)? (we are currently using `Area` not `Areas` e.g. `Area/Admin/Controllers...` - change to `Areas/Admin/Controllers...`)

src/Submissions/Api/Models/Shared/InviteRegisterEntityAdminModel.cs
```diff
@@ -0,0 +1,23 @@
...
+  public string Email { get; set; }
+
+
+  public string EntityName { get; set; } //for label
+  [Required]
+  public string Entity { get; set; }
+  /// <summary>
+  /// The controller to which this will be redirected (either the biobank or ADAC depending on who is adding admins)
+  /// </summary>
+  public string ControllerName { get; set; }
```
beforan, line 22: this will eventually break now that we're adding areas... since controllername strictly speaking won't be enough.

src/Submissions/Api/Program.cs
```diff
@@ -412,6 +412,11 @@
 app.UseHangfireDashboard();
 
 app.MapRazorPages();
+
+app.MapControllerRoute(
+    name: "MyArea",
```
beforan, line 417: `MyArea` is a bit tutorially? I would probably call it `AreasDefault` as it's doing a default catch all for area routes, as opposed to the non area routes below, or specific matches (which we don't have any of...yet)

beforan (changes requested):

src/Submissions/Api/Areas/Admin/Controllers/BiobanksController.cs
```diff
@@ -0,0 +1,211 @@
...
+using Biobanks.Submissions.Api.Utilities;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Http.Extensions;
+using Microsoft.AspNetCore.Identity;
+using Microsoft.AspNetCore.Mvc;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Biobanks.Submissions.Api.Area.Admin.Controllers;
```
beforan, line 18: namespace doesn't match the folders now 😅
```suggestion
namespace Biobanks.Submissions.Api.Areas.Admin.Controllers;
```

src/Submissions/Api/Views/Shared/Emails/ExistingUserRegisterEntityAdminInvite.cshtml
```diff
@@ -0,0 +1,24 @@
+@inject IOptions<SitePropertiesOptions> _config
+@inject IHostEnvironment HostEnvironment
+@{
+	var directoryName = @_config.Value.ServiceName;
+    var emailSignature = @_config.Value.EmailSignature;
+
+	// Format Email Subject
+var emailSubject = HostEnvironment.IsDevelopment()
```
beforan, line 8: this needs to be assigned to `ViewBag.Subject` if you want it to set the subject line.
```suggestion
ViewBag.Subject = HostEnvironment.IsDevelopment()
```

src/Submissions/Api/Views/Shared/Emails/ExistingUserRegisterEntityAdminInvite.cshtml
```diff
@@ -0,0 +1,24 @@
...
+		? $"{directoryName} {ViewBag.Entity} Administration"
+		: $"{directoryName} {ViewBag.Entity} Administration (Environment: {HostEnvironment.EnvironmentName})";
+}
+
+Content-Type: text/plain; charset=utf-8
```
beforan, line 13:
```suggestion
```
we don't do this in the view anymore

src/Submissions/Api/Views/Shared/Emails/NewUserRegisterEntityAdminInvite.cshtml
```diff
@@ -0,0 +1,27 @@
...
+    var emailSignature = @_config.Value.EmailSignature;
+
+	// Format Email Subject
+	var emailSubject = HostEnvironment.IsDevelopment()
+		? $"{directoryName} {ViewBag.Entity} Administration"
+		: $"{directoryName} {ViewBag.Entity} Administration (Environment: {HostEnvironment.EnvironmentName})";
+}
+
+
+Content-Type: text/plain; charset=utf-8
```
beforan, line 16:
```suggestion
```

src/Submissions/Api/Views/Shared/Emails/NewUserRegisterEntityAdminInvite.cshtml
```diff
@@ -0,0 +1,27 @@
+@inject IOptions<SitePropertiesOptions> _config
+@inject IHostEnvironment HostEnvironment
+
+
+@{
+	var directoryName = @_config.Value.ServiceName;
+    var emailSignature = @_config.Value.EmailSignature;
+
+	// Format Email Subject
+	var emailSubject = HostEnvironment.IsDevelopment()
```
beforan, line 10:
```suggestion
	ViewBag.Subject = HostEnvironment.IsDevelopment()
```

src/Submissions/Api/Views/Shared/Emails/NewUserRegisterEntityAdminInvite.cshtml
```diff
@@ -0,0 +1,27 @@
...
+		? $"{directoryName} {ViewBag.Entity} Administration"
+		: $"{directoryName} {ViewBag.Entity} Administration (Environment: {HostEnvironment.EnvironmentName})";
+}
+
+
+Content-Type: text/plain; charset=utf-8
+
+Dear @Html.Raw(ViewBag.Name)
+
+You've been invited to become an admin for @Html.Raw(ViewBag.Entity) at the @directoryName.
```
beforan, line 20: have any of these views been tested? I'm pretty sure 99% of `ViewBag` usage in all the email views should have been changed to Model properties. Because `EmailService` builds a model and assigns values to properties on that model, but the views then never use said model and use ViewBag instead (on which no values have been set!)

beforan: compare to e.g. Decsys (or any of our other projects using this email pattern):

https://github.com/decsys/decsys/blob/main/app/Decsys/Views/Emails/AccountConfirmation.cshtml

there it uses the model and the property on the model.

src/Submissions/Api/Areas/Admin/Controllers/BiobanksController.cs
```diff
@@ -0,0 +1,211 @@
...
+                new
+                {
+                  userId = user.Id,
+                  token = confirmToken
+                },
+                Request.GetEncodedUrl()));
+      }
+      else
+      {
+        return Json(new
```
beforan, line 162: I'm not sure what the `JsonResult`/`Json()` combo are giving us here? they feel a bit .NET Framework MVC ish.

The conventional way to return some JSON in Core is surely just `ActionResult<T>` and then `return objectOfTypeT` or if necessary `IActionResult` and `return Ok(object)`?

Unless that only holds true for ApiControllers and not MVC ones?

src/Submissions/Api/Models/Shared/InviteRegisterEntityAdminModel.cs
beforan: the actual usage of this seems to be in JS, so can we just have a noisy `TODO` to come back and join the dots when we fix the JS urls?

beforan (changes requested):

beforan (approved):

AndyRae (approved):

PR #494, opened 2022-12-08
AB#84331: Requests Controller

MoizIqbal97:
Ports over
- Request Controller from ADAC into admin areas
- Subsquent Models and Views
- Email Sender Service updates with its views


blargly (approved):

src/Submissions/Api/Areas/Admin/Controllers/RequestsController.cs
```diff
@@ -0,0 +1,399 @@
...
+  {
+    _networkService = networkService;
+    _organisationService = organisationService;
+    _userManager = userManager;
+    _tokenLog = tokenLog;
+    _emailService = emailService;
+    _mapper = mapper;
+  }
+
+  public async Task<ActionResult> Requests()
```
beforan, line 45: can't this just be `Index`, so the route is `/Requests` rather than `Requests/Requests`?
```suggestion
  public async Task<ActionResult> Index()
```

beforan: View name might need changing too

beforan (changes requested):

beforan (approved):

blargly (approved):

PR #497, opened 2022-12-13
AB#84312: Analytics controller

AndyRae:
## Overview

Ports the analytics section ADAC to the admin area.

## Notes

Removes the existing API controller in submissions, as it's no longer necessary.


src/Submissions/Api/Areas/Admin/Models/Analytics/DirectoryAnalyticReport.cs
```diff
@@ -0,0 +1,101 @@
...
+        public SearchCharacteristic SearchCharacteristics { get; set; }
+        public EventStat EventStats { get; set; }
+        public ProfilePageStat ProfilePageStats { get; set; }
+
+        public bool IsNullOrEmpty() =>
+            (SessionStats?.IsEmpty() != false) &&
+            (SessionSearchStats?.IsEmpty() != false) &&
+            (SearchCharacteristics?.IsEmpty() != false) &&
+            (EventStats?.IsEmpty() != false) &&
+            (ProfilePageStats?.IsEmpty() != false);
```
dcl10, line 26: I thought the `?` was for nullable variables and none of these appear to be nullable. Also if they can be null you probably want to handle it better than the `?` operator. E.g.
```csharp
SessionStats.HasValue() && !SessionStats.IsEmpty()
```

You would also need to change the declaration above to:
```csharp
public SessionStat? SessionStats ... // and so on...
```

src/Submissions/Api/Areas/Admin/Models/Analytics/DirectoryAnalyticReport.cs
```diff
@@ -0,0 +1,101 @@
...
+        public List<string> AvgNewSessionLabels { get; set; }
+        public List<double> AvgNewSessionCount { get; set; }
+        public List<string> AvgSessionDurationLabels { get; set; }
+        public IList<double> AvgSessionDurationCount { get; set; }
+
+        public bool IsEmpty() => (
+            SessionNumberCount?.Count() +
+            AvgBounceRateCount?.Count() +
+            AvgNewSessionCount?.Count() +
+            AvgSessionDurationCount?.Count())
```
dcl10, line 44: Again here, as above. The `?` is a bit unsafe. They're all lists though and you're checking the counts. So you could keep them us un-nullable, and do this perhaps:
```csharp
public List<string> SessionNumberLabels { get; set; } = new()
// and so on...
```
This would give you some empty lists that, if unfilled, would make your `IsEmpty()` be `false`

src/Submissions/Api/Areas/Admin/Models/Analytics/DirectoryAnalyticReport.cs
```diff
@@ -0,0 +1,101 @@
...
+        public IList<int> SearchTypeCount { get; set; }
+        public IList<string> SearchTermLabels { get; set; }
+        public IList<int> SearchTermCount { get; set; }
+        public IList<string> SearchFilterLabels { get; set; }
+        public IList<int> SearchFilterCount { get; set; }
+
+        public bool IsEmpty() => (
+            SearchTypeCount?.Count() +
+            SearchTermCount?.Count() +
+            SearchFilterCount?.Count())
```
dcl10, line 60: Same as above

src/Submissions/Api/Areas/Admin/Models/Analytics/DirectoryAnalyticReport.cs
```diff
@@ -0,0 +1,101 @@
...
+        public IList<int> ContactNumberCount { get; set; }
+        public IList<string> FilteredContactLabels { get; set; }
+        public IList<int> FilteredContactCount { get; set; }
+        public IList<string> FilteredMailToLabels { get; set; }
+        public IList<int> FilteredMailToCount { get; set; }
+
+        public bool IsEmpty() => (
+            ContactNumberCount?.Count() +
+            FilteredContactCount?.Count() +
+            FilteredMailToCount?.Count())
```
dcl10, line 76: Same as above

src/Submissions/Api/Areas/Admin/Models/Analytics/DirectoryAnalyticReport.cs
```diff
@@ -0,0 +1,101 @@
...
+
+    public class ProfilePageStat
+    {
+        public IList<SourceCount> ProfileSources { get; set; }
+        public IList<string> PageRouteLabels { get; set; }
+        public IList<int> RouteCount { get; set; }
+
+        public bool IsEmpty() => (
+                ProfileSources?.Count() == 0 ||
+                (ProfileSources?.Count() == 1 &&
```
dcl10, line 88: same as above

dcl10 (changes requested):

MoizIqbal97 (approved):

PR #505, opened 2023-01-11
AB#85178: Profile controller for network area

MoizIqbal97:
Ports over profile section from the network controller
Ports over models and views
Removing and commenting out session related code


src/Submissions/Api/Extensions/HttpPostedFilebaseExtensions.cs
```diff
@@ -0,0 +1,43 @@
+using Microsoft.AspNetCore.Http;
+using System;
+using System.IO;
+using System.Linq;
+
+namespace Biobanks.Submissions.Api.Extensions;
+
+public static class HttpPostedFilebaseExtensions
```
AndyRae, line 8: Should rename this class / file - as we're now extending `IFormFile` and not `HttpPostedFilebase`

AndyRae (changes requested):

blargly (approved):

AndyRae (approved):

PR #504, opened 2023-01-11
AB#76211: Register controller

AndyRae:
## Overview

Ports the `RegisterController` from directory to core. 

Includes:
- Migrates the `VerifyRecaptchaAttribute` filter in directory to a new `RecaptchaService`. This isn't specific to registration, and can be reused on other forms.

PR #506, opened 2023-01-12
AB#85109: Profile Controller Update

MoizIqbal97:
Program.cs dependency injection update


beforan (changes requested):

src/Submissions/Api/Program.cs
```diff
@@ -203,13 +205,15 @@
     .AddTransient<ICollectionService, CollectionService>()
     .AddTransient<IConfigService, ConfigService>()
     .AddTransient<IContentPageService, ContentPageService>()
+    .AddTransient<IEmailService,EmailService>()
```
AndyRae, line 208: This is actually already registered, done in an extension method - `Submissions/Api/Extensions/ServiceCollectionExtensions.cs`
I literally updated it in the last PR #504, so if it wasn't registering properly it now should.

It's done in this extension method as it's ported from Simun / Sarg

AndyRae (changes requested):

beforan (approved):

AndyRae (approved):

PR #509, opened 2023-01-16
AB#84952: Replace MVP sitemap - breadcrumbs

AndyRae:
## Overview

Functionality to replace MVPSitemap in the new core app. Uses the `cloudscribe.Web.Navigation` library to generate breadcrumbs on views (see image).

Screenshot 2023-01-16 at 16 11 59

## Bonus
- [Move `PagesAdminController`](https://github.com/biobankinguk/biobankinguk/commit/94e3983dd1139c2f36b73caa6fe25017feb9bd34) from submissions to directory.
- Fix `TermController` to enable public use, and how it retrieved data in the `CollectionService`.
- Adds `EditorTemplates` to Register pages, see image for form update.


src/Submissions/Api/Controllers/Directory/PagesAdminController.cs
```diff
@@ -43,6 +43,10 @@ public async Task<ActionResult> ContentPage(string slug)
             {
                 if (page.IsEnabled || User.IsInRole("ADAC"))
                 {
+                    // Set the last breadcrumb
+                    var breadCrumbHelper = new TailCrumbUtility(HttpContext);
```
beforan, line 47: what are these actions doing to the breadcrumbs, and why?

do all controller actions need to manually set breadcrumbs now?

I assume not since they've not all been updated. but in that case, why these ones?

AndyRae: Controller actions do not need to manually set breadcrumbs, this is still derived from the xml sitemap as before.

There were 2 existing controllers that set overrides to the breadcrumbs, specifically the ContentPages (to set the page title as the breadcrumb), and the CollectionController (which set the sample set collection ID as it).

Effectively, this line in the framework app did this.
`SiteMaps.Current.CurrentNode.ParentNode.RouteValues["id"] = sampleSet.CollectionId;`
or through the sitemap annotation: `[SiteMapTitle("Title")]`

Has turned into setting them manually through the [TailCrumbUtility](https://www.cloudscribe.com/adding-extra-breadcrumbs).

beforan: cool, thanks for aiding my understanding 👍

beforan (approved):

blargly (approved):

PR #523, opened 2023-01-26
AB#92911: Seeding config

MoizIqbal97:
Seeds config data into Config table 

image


beforan:
#517 did this work as part of introducing `FixedRefDataSeeder`

PR #528, opened 2023-01-31
AB#89361: Admin Funders Controller testing

AndyRae:
## Overview

Fixes for the Admin area Funders Controller, after testing.

Includes:
- Fixing the controller so CRUD operations work, with feedback in the right places. (images below)
- Fix the Navbar links so they work inside and outside of areas.
- Fix ADACTabs,  which needs to load as `@await Html.PartialAsync` to work. (background, last image) All other views will need this change also.

Screenshot 2023-01-31 at 15 53 20

Screenshot 2023-01-31 at 15 56 11

Screenshot 2023-01-31 at 15 56 36

Screenshot 2023-01-31 at 15 56 46

Screenshot 2023-01-31 at 16 35 04

Screenshot 2023-01-31 at 16 58 12


MoizIqbal97 (approved):

beforan (approved):
slightly weird and hilarious that the delete warning isn't a modal, when so many other things are 😅

PR #530, opened 2023-02-01
AB#89381: Allow Block List

MoizIqbal97:
Changes to fix Admin/RegistrationDomainRules/BlockAllowList

image


src/Submissions/Api/Frontend/js/pages/Admin/adac-blockallow-list.js
```diff
@@ -1,139 +1,157 @@
...
+
+  $(".delete-confirm").click(function (e) {
+    e.preventDefault();
+
+    var $link = $(this);
+    var linkData = $link.data("domain-rule");
+    var url = $link.data("resource-url") + "/" + linkData.Id;
+
+    bootbox.confirm(
+      "Are you sure you want to delete " + linkData.Value + "?",
```
AndyRae, line 107: Based on fixing the ref data patterns:

Does this work? (does the modal show the value of the rule?)
Fix is probably to lowercase if not:
```suggestion
      "Are you sure you want to delete " + linkData.value + "?",
```

And does it actually delete successfully? If not, again fix is lowercase on the linkData.id (line 104)

src/Submissions/Api/Startup/Web/ConfigureWebServices.cs
```diff
@@ -217,6 +217,7 @@ public static WebApplicationBuilder ConfigureServices(this WebApplicationBuilder
       var xmlFilename = $"swagger.xml"; //File name should match the documation being called in Api.csproj
       opts.IncludeXmlComments(Path.Combine(AppContext.BaseDirectory, xmlFilename));
     });
+    AppContext.SetSwitch("Npgsql.EnableLegacyTimestampBehavior", true);
```
beforan, line 220: this is done globally in another PR, so i think can be removed from here
```suggestion
```

beforan (changes requested):

AndyRae (approved):

PR #534, opened 2023-02-03
AB#89387: Admin Registration Requests testing

AndyRae:
## Overview

Fixes the Admin area Requests controller, and the registration user journeys.

Mostly fixing the routing with areas, and a few email gotchas.

This is now working - users can register a request, have it approved, be sent the relevant emails, with correct links to activate their account.

### DIFF
--- a/src/Submissions/Api/Areas/Admin/Controllers/RequestsController.cs
+++ b/src/Submissions/Api/Areas/Admin/Controllers/RequestsController.cs
@@ -139,10 +139,11 @@ public class RequestsController : Controller
           Url.Action("Confirm", "Account",
               new
               {
+                Area = "",
                 userId = user.Id,
                 token = confirmToken
               },
-              Request.GetEncodedUrl())
+              Request.Scheme)
           );
     }
     else
@@ -152,9 +153,10 @@ public class RequestsController : Controller
           new EmailAddress(request.UserEmail),
           request.UserName,
           request.OrganisationName,
-          Url.Action("SwitchToBiobank", "Account",
+          Url.Action("Edit", "Profile",
               new
               {
+                Area = "Biobank",
                 id = request.OrganisationRegisterRequestId,
                 newBiobank = true
               },
--- a/src/Submissions/Api/Areas/Admin/Controllers/RequestsController.cs
+++ b/src/Submissions/Api/Areas/Admin/Controllers/RequestsController.cs
@@ -1,13 +1,11 @@
 using AutoMapper;
 using Biobanks.Data.Entities;
-using Biobanks.Submissions.Api.Areas.Admin.Models;
 using Biobanks.Submissions.Api.Constants;
 using Biobanks.Submissions.Api.Models.Emails;
 using Biobanks.Submissions.Api.Services.Directory.Contracts;
 using Biobanks.Submissions.Api.Services.Directory.Dto;
 using Biobanks.Submissions.Api.Services.EmailServices.Contracts;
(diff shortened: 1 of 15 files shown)


src/Submissions/Api/Areas/Admin/Controllers/RequestsController.cs
```diff
@@ -140,10 +137,11 @@ public async Task<ActionResult> AcceptBiobankRequest(int requestId)
           Url.Action("Confirm", "Account",
               new
               {
+                Area = "",
                 userId = user.Id,
                 token = confirmToken
               },
-              Request.GetEncodedUrl())
+              Request.Scheme)
```
AndyRae, line 144: Essentially this is fetching the request protocol (`https`) - `getEncodedUrl` mistakenly fetched the entire url and sent a broken link.

src/Submissions/Api/Views/Shared/Emails/ResendConfirm.cshtml
```diff
@@ -0,0 +1,28 @@
+@inject IOptions<SitePropertiesOptions> _config
+@inject IHostEnvironment HostEnvironment
+
+@{
+  var directoryName = @_config.Value.ServiceName;
+
+	// Format Email Header {"<Name>" <Address>}
+	var emailFrom = String.Format(@"""{0}"" {1}",
```
beforan, line 8: this may not be necessary.

if it is, please use string interpolation over string.format if at all possible

src/Submissions/Api/Views/Shared/Emails/ResendConfirm.cshtml
```diff
@@ -0,0 +1,28 @@
...
+	  directoryName,
+	  @_config.Value.ContactAddress
+		);
+
+	// Format Email Subject
+	var emailSubject = HostEnvironment.IsDevelopment()
```
beforan, line 14: 90% sure this ternary is the wrong way round.

and actually, `IsDevelopment` is the wrong condition.

The desirable behaviour is that all non-prod environments include their environment.

It occurs to me it'd be much better to do this append in the Email Service somewhere (possibly even Email Sender) rather than in every single email.

beforan: (and, to be clear, prod should NOT include an environment in its subject line as that's kind of unprofessional leakage to customers)

beforan (changes requested):

src/Submissions/Api/Views/Shared/Emails/ResendConfirm.cshtml
beforan: actually, SargAssure's EmailSender does include this functionality: https://github.com/SargAssure/SargAssure/blob/6ceada559151dabb80f54b297c81375a38077cd8/app/SargAssure/Services/EmailSender/SendGridEmailSender.cs#L32

beforan: and so does SIMUN

blargly (approved):

src/Submissions/Api/Views/Shared/Emails/ResendConfirm.cshtml
AndyRae: Have removed this from all the emails templates - the email service does include this as well.

beforan (approved):

blargly (approved):

PR #535, opened 2023-02-06
AB#89392: Admin SiteConfig testing

AndyRae:
## Overview

Fixes after testing the Admin area, SiteConfig controller.

No big changes, just adding JS script tags to the views, including markdown, and small bits.

This portion is now working, with the ability to set configuration site wide.

Screenshot 2023-02-06 at 11 26 18

---

Screenshot 2023-02-06 at 11 22 34

---

Screenshot 2023-02-06 at 11 22 45

---

Screenshot 2023-02-06 at 11 22 21

### DIFF
--- a/src/Submissions/Api/bundleconfig.json
+++ b/src/Submissions/Api/bundleconfig.json
@@ -81,10 +81,20 @@
     "outputFileName": "wwwroot/dist/js/markdown-editor.min.js",
     "inputFiles": [
       "Frontend/node_modules/markdowndeep/clientSide/MarkdownDeepLib.min.js", // TODO: copy images
-      "Frontend/js/init/markdown-editor.js" // TODO: copy help file from the package
+      "Frontend/js/init/markdown-editor.js"
     ],
     "minify": { "enabled": true },
     "sourceMap": true
+  },
+
+  // Markdown Deep
+  {
+    "outputFileName": "wwwroot/dist/mdd_help.htm",
+    "inputFiles": [
+      "Frontend/node_modules/markdowndeep/clientSide/mdd_help.htm"
+    ],
+    "minify": { "enabled": false },
+    "sourceMap": false
   },
 
   // Page bundles (grouped by Controller)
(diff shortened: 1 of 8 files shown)


src/Submissions/Api/bundleconfig.json
```diff
@@ -81,12 +81,22 @@
     "outputFileName": "wwwroot/dist/js/markdown-editor.min.js",
     "inputFiles": [
       "Frontend/node_modules/markdowndeep/clientSide/MarkdownDeepLib.min.js", // TODO: copy images
```
beforan, line 83: Do the images still need copying? did the toolbars appear correctly without doing manual copying?

AndyRae: The toolbars work as pictured, without manual copying (I think this is a backup toolbar maybe?). It's difficult to tell, as I can't see a working version, but this is obviously not the full colour toolbar that the library _probably_ wants to load, which needs to be added manually. 

The functionality/UX works fine, but to be perfect; the toolbar, a background vector (though unused), and a loading GIF need to be copied manually. 
Problem is, the library specifically wants them to be in `dist/css/` (which is where the bundler is bundling to, so don't want to remove `dist` from `.gitignore`).
The itself bundler cannot process images unfortunately.

So at a kind of impasse without changing the bundle directory, or where this specific library is bundled to. But the idea is to handle this with the other frontend bundling that needs to looked at in [#75748](https://dev.azure.com/UniversityOfNottingham/DRS/_backlogs/backlog/ADAC0001%20-%20Biobanks/Epics/?workitem=75748) (fonts, other images).

blargly (approved):

src/Submissions/Api/bundleconfig.json
beforan: 1. i don't ever recall a full colour toolbar used in the old app 😅
2. yes any copying of assets can be done the same way as fonts etc. we shouldn't need to commit anything in dist; it's accessible for the csproj to configure msbuild to copy stuff into there at build time if appropriate

beforan (approved):

PR #559, opened 2023-03-08
AB#99140: Breadcrumbs missing

MoizIqbal97:
Added missing breadcrumbs in navigation


dcl10 (approved):

src/Submissions/Api/navigation.xml
```diff
@@ -162,13 +165,11 @@
 
     <!-- Content Pages -->
     <NavNode key="AdminRequests" text="Admin" controller="Requests" area="Admin" action="Index">
```
AndyRae, line 167: Think you can remove this one altogether - it's a duplicate from higher up and isn't really for this section.

Issue #606, opened 2024-08-23
🐛 Connection pooling results in app crashing


AndyRae:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The app crashes periodically after some use. The only fix is to restart the app, and often have to leave it turned off for a short time. 

### Expected Behavior

It should not crash.

### Steps To Reproduce

Have the app running in a deployed environment - presumably with normal workload.

### Environment

```markdown
- OS: Linux
- Other environment details: Azure App Service
```

### Anything else?

Logs:
```
Npgsql.NpgsqlException (0x80004005): The connection pool has been exhausted, either raise 'Max Pool Size' (currently 100) or 'Timeout' (currently 15 seconds) in your connection string.
```

Indicates that there is a connection leak somewhere in the application code.

AndyRae:
Further to this, inspecting the logs for TDCC, indicates that the site is under continuous search requests, sometimes these requests increase in volume and cause the above error.
These requests are continually polling different permutations of the search facets for search terms.
There may be help in profiling the search mechanisms to check for connection leaks... (I don't have a good understanding of this yet).

But a theory is that this is bots indexing the search terms, as every term is actually a hyperlink in the html of the site, even adding facets etc. At some points this indexing gets swarm like, and as there is no caching etc, every request is a new db connection not returned to the pool. I might point to the noticeable presence for the [search metadata on Google](https://www.google.com/search?q=https%3A%2F%2Fdirectory.biobankinguk.org%2FSearch%2FCollections&sca_esv=3bd69d27f24c6fbe&sca_upv=1&source=hp&ei=CrDQZpuhLbG4hbIP8-78-A4&iflsig=AL9hbdgAAAAAZtC-GpBcQ9ZEdmx5jgGQSQIhGg4aImUZ&ved=0ahUKEwibzr-R3ZqIAxUxXEEAHXM3H-8Q4dUDCBk&uact=5&oq=https%3A%2F%2Fdirectory.biobankinguk.org%2FSearch%2FCollections&gs_lp=Egdnd3Mtd2l6IjVodHRwczovL2RpcmVjdG9yeS5iaW9iYW5raW5ndWsub3JnL1NlYXJjaC9Db2xsZWN0aW9uc0i3AlAAWABwAHgAkAEAmAFRoAFRqgEBMbgBA8gBAPgBAvgBAZgCAKACAJgDAJIHAKAHaQ&sclient=gws-wiz) as an example. 
[...]

Issue #580, opened 2024-01-12
When creating a collection, disease term selection sometimes stores the incorrect term


beforan:
## Environment details

- Operating System: unknown
- Browser: unknown
- Directory environment: biobankinguk prod

## Repro steps:

- Create a new collection
- Choose the disease term "Malignant tumor of pancreas (disorder)"
- Fill out the rest of the collection and save
- The saved collection has the disease term "Local recurrence of malignant tumor of pancreas (disorder)"

## Acceptance criteria

- The term stored should match the term selected at creation

## Status

- [ ] Reproduction confirmed

## Further information

From a preliminary investigation into the environment's data, this isn't an "other terms" issue (I initially thought it might be that the user was selecting an "other term" at creation, and then the collection was showing the "canonical" term). In this case, there are two different terms in play: one enabled for gui use (`DisplayOnDirectory`), the other not. So literally a different SNOMED id is stored than the one that was selected.

I do wonder if we're matching incorrectly on lowercase string values - the values in the table are mixed case and so "Malignant" (the selected term) and "[...] malignant" (the stored term). This needs to be investigated.

beforan:
## Initial findings

This is definitely due to case insensitive matching, resulting in finding multiple matches despite the whole term being used, then selecting the incorrect "first" match from those.

This is an issue in production due to the collation in use on the production SQL Server database (allowing case insensitivity)

It's not an issue that appears when running  v4.0.0 (ASP.NET Core version) of the Directory so far, as we target Postgres and the default postgres collation is case sensitive. This however has caused other issues to appear.

## Initial fix

We have a locally working fix that relies on the postgres case sensitive collation to resolve this issue, and have resolved the other issues caused when case insensitivity *is* desirable.

## Better fix

While the above fixes the reported scenario, it's possible that even with case-sensitive matching, we may erroneously pick an incorrect disease status while we use the partial match code for what should be an exact match. If it's not too hard to isolate the times when this should be done, it would be best to provide up to 3 different Disease Status lookup methods:

1. Case insensitive partial match (as used today by Search and other typeahead scenarios)
2. Case sensitive exact match (as should be used by this scenario, and should select a Single result, not the First of potentially many)
3. Optionally case sensitive partial match (if there is a need for this - all scenarios may actually be covered by the above 2)

[...]

PR #607, opened 2024-08-31
AB#157605: Numerous Fixes 

### LINKED ISSUE #598: Historical tab
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When I go to the historical tab as the admin user, it does not show recent decisions. 

https://directory.biobankinguk.org/ADAC/Historical

### Expected Behavior

It would list the biobanks we have said yes/no to

### Steps To Reproduce

Register a new biobank
Accept/reject the request
It wont appear here: https://directory.biobankinguk.org/ADAC/Historical

### Environment

```markdown
- OS:
- Other environment details:
```

### Anything else?

_No response_

### LINKED ISSUE #605: 🐛 Register biobanks with inconsistent cased user emails is invisible
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When registering a new biobank as a user, if you use a different case in your email to the user you are registered as, you will not be granted permission to access that biobank, to then complete registration.

### Expected Behavior

I would expect a user who has registered a new biobank, and had it approved, be able to then complete registration.

### Steps To Reproduce

1. Have a user with `test@example.com`
2. Register a new biobank, logged in or out it doesn't matter.
3. Register the biobank admin as `TEST@example.com`
4. Approve the registration as the Directory Admin.
5. `test@example.com` will be logged out, and they will receive an email that is has been approved.
6. Click the link
7. Log in as `test@example.com`
8. This will 401, and you will not see the biobank to register in the dropdown.

### Environment

```markdown
- OS:
- Other environment details:
```

### Anything else?

The user is added the correct permissions in the Biobank register controller. And the request persists, that specific email just doesn't have access to see it.

The problem is when adding permissions to the user when they log in. Inside `CustomClaimsPrincipalFactory` the claims are collected for biobanks, and registrations - it calls `ListAcceptedRegistrationRequestsByUserId`. 
https://github.com/biobankinguk/biobankinguk/blob/5fa2686564da52245de53d2633c9d31088964b31/app/Directory/Services/Directory/OrganisationDirectoryService.cs#L407

This filters by an email address (case sensitive) on the register request. The fix here is - get the normalized email instead, and additionally normalise the email in the filtering. 

So a user might use a different case in their email as for when they registered originally, and it will therefore not attach the permissions to them.
The code bug existed in the framework code also, but might not have been present in the UX - as I think permissions were built differently and not using claims in the same way.

AndyRae:
## Overview

A series of small fixes for production issues, and bumps the version to 4.0.0

Closes #605 - Adding filtering by normalised emails to build claims.
Closes #598 - Adding default ordering by date descending, and fixing the serialized date format.
Relates #580 - Adding an exact match lookup for Ontologies, this fixes adding collections with similar names.
Relates #606 - Adds a `robots.txt`, and adds search `nofollow` instructions for robots. (we will see if this helps reduce robots for now)
And removes the exclamation mark to indicate actions as a dropdown menu for biobanks - it's too messy to try and do this for now as the menu is fetched from Wordpress.

### DIFF
--- a/app/Directory/Areas/Admin/Views/Requests/Historical.cshtml
+++ b/app/Directory/Areas/Admin/Views/Requests/Historical.cshtml
@@ -44,7 +44,7 @@
 							@request.Action
 						</td>
 						<td>
-							@request.Date
+							@request.Date.ToString("dd/MM/yyyy HH:mm:ss")
 						</td>
 
 
--- a/app/Directory/Areas/Admin/Controllers/RequestsController.cs
+++ b/app/Directory/Areas/Admin/Controllers/RequestsController.cs
@@ -463,7 +463,7 @@ public class RequestsController : Controller
 
       var model = new HistoricalModel
       {
-          HistoricalRequests = bbRequests.Concat(nwRequests).ToList()
+          HistoricalRequests = bbRequests.Concat(nwRequests).OrderByDescending(x => x.Date).ToList()
       };
 
       return View(model);
--- a/app/Directory/Areas/Biobank/Controllers/CollectionsController.cs
+++ b/app/Directory/Areas/Biobank/Controllers/CollectionsController.cs
@@ -191,7 +191,7 @@ public class CollectionsController : Controller
           })
           .ToList();
 
-      var ontologyTerm = await _ontologyTermService.Get(value: model.Diagnosis);
+      var ontologyTerm = await _ontologyTermService.Get(value: model.Diagnosis, doExactMatch: true);
 
       // Create and Add New Collection
       var collection = await _collectionService.Add(new Collection
--- a/app/Directory/Auth/CustomClaimsPrincipalFactory.cs
+++ b/app/Directory/Auth/CustomClaimsPrincipalFactory.cs
@@ -58,7 +58,6 @@ public class CustomClaimsPrincipalFactory : UserClaimsPrincipalFactory<Applicati
         .Where(x => !identity.HasClaim(c => c.Type == x.Type && c.Value == x.Value));
       
       var biobankRequestClaims = organisationsRequests
-        .Where(x => x.UserEmail == user.Email)
         .Select(x => new KeyValuePair<int, string>(x.OrganisationRegisterRequestId, x.OrganisationName))
         .Select(x => new Claim(CustomClaimType.BiobankRequest, JsonSerializer.Serialize(x)))
         .Where(x => !identity.HasClaim(c => c.Type == x.Type && c.Value == x.Value));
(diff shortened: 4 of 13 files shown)


Jivanam (approved):

Dev-Akashili (approved):

AndrewThien (approved):

app/Directory/Areas/Admin/Views/Requests/Historical.cshtml
```diff
@@ -44,7 +44,7 @@
 							@request.Action
 						</td>
 						<td>
-							@request.Date
+							@request.Date.ToString("dd/MM/yyyy HH:mm:ss")
```
beforan, line 47: theoretically datatables js would normally be doing the visual formatting on something like this. Not a problem doing it this way but just a note around how the codebase did things.

beforan (approved):

app/Directory/Areas/Admin/Views/Requests/Historical.cshtml
AndyRae: I thought so! As in this line?

https://github.com/biobankinguk/biobankinguk/blob/5fa2686564da52245de53d2633c9d31088964b31/app/Directory/Frontend/js/pages/Admin/adac-historical.js#L2

My problem is that when deployed the dates would still be mm/dd/yyyy in the deployed instance, but locally work fine as dd/mm/yyyy. So quick fix is clearly to be explicit here instead, but maybe there is something not working with the Javascript in deployment.

beforan: Yeah, my main thought to flag it was in case it was an indicator of js not working as expected more broadly.

Obviously no need for it to halt this PR and this specific fix, but might be worth backlogging/investigating further.

beforan: > I thought so! As in this line?
> 
> https://github.com/biobankinguk/biobankinguk/blob/5fa2686564da52245de53d2633c9d31088964b31/app/Directory/Frontend/js/pages/Admin/adac-historical.js#L2
> 
> when deployed [...] mm/dd/yyyy [...], but locally [...] dd/mm/yyyy.

wonder if the js is working but it's a locale/culture issue? e.g. azure servers will mostly likely be US configured. but since it's doing this client side I may be wrong. you'd like to think it'd use browser culture, but depending on the datatables plugin/config, that'd explain different formats between local and deployed.

AndyRae: Yes other bits of the JS do indicate that it's working fine - I did look at the locale/culture, but it's really not clear to me how to override it to me, the few config settings I tried didn't work. 
And only a brief look, but this is the only serialized date I can find also, but would be nice to check this behaviour.

PR #611, opened 2024-11-15
AB#186702 Added SMTP email service

blargly:
## Overview

Adding SMTP email option. Tested in other code bases.

Updated a couple of vulnerable packages.


brian-kim31 (dismissed):

app/Directory/Services/EmailSender/SmtpEmailSender.cs
Jivanam, line 1: I suggest using the one from `AI4Green4Students` as there was disposal issues. E.g. when calling `SendEmail` in a loop.

prquinlan:
Given the time passed, close?

